PR #14, opened 2019-06-26
First version of Russian translation

unhappytroll:
First version of Russian translation, for testing purposes, will be corrected in the future


addons/common/stringtable.xml
```diff
@@ -68,105 +68,139 @@
...
             <English>Aircraft</English>
+            <Russian>Самолёт</Russian>
         </Key>
         <Key ID="STR_ZEN_Common_Filter">
             <English>Filter</English>
+            <Russian>Фильтр</Russian>
         </Key>
         <Key ID="STR_ZEN_Common_Units">
             <English>Units</English>
+            <Russian>Солдаты</Russian>
```
CreepPork: It's not only soldiers, could be just some unit (inc. civilian side)

unhappytroll: that depends on particular point of application. we have quite a bunch of synonyms, which can be applied under that word.

addons/attributes/stringtable.xml
```diff
@@ -1,18 +1,21 @@
-<?xml version="1.0" encoding="utf-8"?>
```
mharis001: What changed in these lines?

addons/editor/stringtable.xml
```diff
@@ -1,58 +1,69 @@
...
         </Key>
         <Key ID="STR_ZEN_Editor_DeclutterEmptyTree">
             <English>Declutter Empty Tree</English>
+            <Russian>Очистить пустое дерево</Russian>
             <German>Leer Kategorie aufräumen</German>
             <Polish>Oczyść drzewo pustych obiektów</Polish>
         </Key>
         <Key ID="STR_ZEN_Editor_DeclutterEmptyTree_Description">
             <English>Declutters the empty tree by removing empty vehicles as those are available in the side specific trees</English>
+            <Russian>???</Russian>
```
mharis001: If you are unsure about the translation, leave empty.

addons/attributes/stringtable.xml
CreepPork: Line endings probably.

mharis001: BOM character, fixed.

mharis001:
Merging to avoid any further conflicts.

PR #33, opened 2019-06-29
Allow building with armake

CreepPork:
- Remove unused includes from `garage, inventory, vision` components
- Make armake compile without throwing missing file errors
- Fix armake not processing `GUI_BCG_COLOR` macro properly, so it has now been renamed to `GUI_THEME_COLOR`
- Fixed `make.py` not compiling in `modules` due to trailing `\` (seems to affect only my machine, though 🤔)
- [x] `setup.py` fails as the includes it moves don't exist


.gitignore
```diff
@@ -8,3 +8,5 @@ texHeaders.bin
 *.biprivatekey
 Thumbs.db
 CHANGELOG.md
+.hephaestus.json
```
jonpas: What's this?

CreepPork: Special config file for [Hephaestus](https://github.com/ArmaAchilles/Hephaestus).
It doesn't have to be in source control as it has full sys paths. I personally use it to quickly start the game and build using armake.

jonpas: Local gitignore?

Add HEMTT support btw. :D

CreepPork: I'll add CI support after this PR and then I'll start adding HEMTT support.

CreepPork: I'm probably not going to be the only one using it, so I'll just leave it for now.

PR #35, opened 2019-06-29
Fix cycle waypoint position

oOKexOo:
- Fix https://feedback.bistudio.com/T85817

### DIFF
--- a/addons/editor/functions/fnc_bi_curatorWaypointPlaced.sqf
+++ b/addons/editor/functions/fnc_bi_curatorWaypointPlaced.sqf
@@ -32,9 +32,9 @@ if (GVAR(unitRadioMessages) != 2) then {
 };
 
 if (_waypointID == 1) then {
-    [_group, 0] setWaypointPosition [getPos _leader, 0];
+    [_group, 0] setWaypointPosition [getPosASL _leader, -1];
 };
 
 if (_waypointType == "CYCLE") then {
-    _waypoint setWaypointPosition [waypointPosition [_group, 0], 0];
+    _waypoint setWaypointPosition [AGLToASL waypointPosition [_group, 0], -1];
 };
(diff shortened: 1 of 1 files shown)


addons/editor/functions/fnc_bi_curatorWaypointPlaced.sqf
```diff
@@ -1,7 +1,7 @@
 /*
- * Author: Bohemia Interactive, mharis001
+ * Author: Bohemia Interactive, mharis001, Kex
  * Handles placement of a waypoint by Zeus.
- * Edited to allow control over radio messages.
+ * Edited to allow control over radio messages and fixing waypoint positioning.
```
CreepPork:
```suggestion
 * Edited to allow control over radio messages and fixed cycle waypoint position.
```

addons/editor/functions/fnc_bi_curatorWaypointPlaced.sqf
```diff
@@ -18,11 +18,23 @@
...
+
+if (GVAR(unitRadioMessages) != 2) then {
+    [
+        _leader,
+        ["CuratorWaypointPlaced", "CuratorWaypointPlacedAttack"] select (_waypointType == "DESTROY")
+    ] call BIS_fnc_curatorSayMessage;
+};
+
+if (_waypointID == 1) then {
+    [_group, 0] setWaypointPosition [getPos _leader, 0];
```
mharis001: `getPos` returns AGLS, `setWaypointPosition` expects AGL or Object. Can pass the leader directly here.

Do we want to force exact position here? If so, then this should use ASL with a -1 radius.

addons/editor/functions/fnc_bi_curatorWaypointPlaced.sqf
```diff
@@ -18,11 +18,23 @@
...
+};
+
+if (_waypointType == "CYCLE") then {
+    _waypoint setWaypointPosition [waypointPosition [_group, 0], 0];
```
mharis001: Should this also force exact position with -1 as the radius?

addons/editor/functions/fnc_bi_curatorWaypointPlaced.sqf
oOKexOo: Yes, let's go for the -1 option.

PR #36, opened 2019-06-29
Fix copy/pasted modules

oOKexOo:
- Fix activation of copy/pasted modules (_e.g._ smoke shell)


addons/editor/functions/fnc_bi_curatorObjectPlaced.sqf
```diff
@@ -28,6 +29,11 @@ if (GVAR(unitRadioMessages) == 0 && {!isNull _group && {side _group in [west, ea
     [effectiveCommander _object, "CuratorObjectPlaced"] call BIS_fnc_curatorSayMessage;
 };
 
+if (_object isKindOf "Module_f") then
+{
```
jonpas: On same line.

addons/editor/functions/fnc_bi_curatorObjectPlaced.sqf
```diff
@@ -28,6 +29,11 @@ if (GVAR(unitRadioMessages) == 0 && {!isNull _group && {side _group in [west, ea
...
+	_object setVariable ["BIS_fnc_initModules_activate", true, true];
```
jonpas: Wrong indent.

addons/editor/functions/fnc_bi_curatorObjectPlaced.sqf
```diff
@@ -28,6 +29,10 @@ if (GVAR(unitRadioMessages) == 0 && {!isNull _group && {side _group in [west, ea
     [effectiveCommander _object, "CuratorObjectPlaced"] call BIS_fnc_curatorSayMessage;
 };
 
+if (_object isKindOf "Module_f") then {
```
mharis001:
```suggestion
if (_object isKindOf "Module_F") then {
```

mharis001:
Since ZEN modules do not rely on BI's module framework but still inherit from `Module_F`, I think this should also check: `!(_object isKindOf QGVAR(moduleBase)` to reduce the number variable broadcasts needed. And maybe move this to the modules component?

oOKexOo:
Ok, I could add a new class EH to the curator module for attaching a `CuratorObjectPlaced` EH, which in return calls something like `fnc_curatorModulePlaced.sqf`. All located in module component ofc.

addons/modules/XEH_preInit.sqf
```diff
@@ -6,4 +6,15 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+["ModuleCurator_F", "Init", {
+    params ["_logic"];
+
+    _logic addEventHandler ["CuratorObjectPlaced",
+    {
```
mharis001: Same line.

addons/modules/XEH_preInit.sqf
```diff
@@ -6,4 +6,15 @@ PREP_RECOMPILE_START;
...
+        if (param[1] isKindOf "Module_F") then {
```
mharis001: Use params. `params ["", "_object"];`

addons/modules/XEH_preInit.sqf
```diff
@@ -6,4 +6,15 @@ PREP_RECOMPILE_START;
...
+            _this call FUNC(handleModulePlaced);
```
mharis001: Don't think we need a separate function for this. Can just do the `&& {!(_logic isKindOf QGVAR(moduleBase))}` check in the if statement above.

addons/modules/XEH_preInit.sqf
```diff
@@ -6,4 +6,15 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+["ModuleCurator_F", "Init", {
```
mharis001: Comment above this line maybe?
`// Fix copy/pasted modules that use BI's module framework`

addons/modules/XEH_preInit.sqf
oOKexOo: Damn, still need to get used to K&R style.

PR #42, opened 2019-07-01
Add skills button to attributes displays

mharis001:
- Add skills button to unit and group attributes displays to allow for fine-tuning of AI sub-skills
- Add `SLIDER:PERCENT` sub-type to the slider dialog control

### DIFF
--- a/addons/attributes/RscAttributes.hpp
+++ b/addons/attributes/RscAttributes.hpp
@@ -887,6 +887,10 @@ class GVAR(RscAttributesMan): GVAR(RscAttributesBase) {
             text = "$STR_A3_Arsenal";
             function = QFUNC(buttonArsenal);
         };
+        class Skills {
+            text = CSTRING(Skills);
+            function = QFUNC(buttonSkills);
+        };
     };
 };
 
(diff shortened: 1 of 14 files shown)


addons/attributes/RscAttributes.hpp
```diff
@@ -1057,6 +1061,12 @@ class GVAR(RscAttributesGroup): GVAR(RscAttributesBase) {
         class ButtonOK: ButtonOK {};
         class ButtonCancel: ButtonCancel {};
     };
+    class buttons {
```
CreepPork:
```suggestion
    class Buttons {
```

I'm not certain how arma reacts to this, but it looks better this way.

addons/attributes/functions/fnc_buttonSkills.sqf
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Example:
+ * [CONTROL] call zen_attributes_fnc_buttonSkills
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+private _entity = GETMVAR(BIS_fnc_initCuratorAttributes_target,objNull);
+private _unit   = if (_entity isEqualType grpNull) then {leader _entity} else {_entity};
```
CreepPork: Maybe a select?

addons/attributes/RscAttributes.hpp
mharis001: The config is not case-sensitive, I left it lowercase to match the other ones in display configs: `controls` and `controlsBackground`, but I agree it looks better capitalized 😄. Will change the others.

addons/attributes/functions/fnc_buttonSkills.sqf
mharis001: This is mostly personal preference, but `ARRAY select BOOL` does require building the array first. I think the rest of this component uses `if else` for attributes that work with multiple entity types.

PR #45, opened 2019-07-02
Add traits button to attributes display

mharis001:
- Add traits button to unit attributes display - medic, engineer, eod
- Compatible with ACE medical, repair, and  explosives


addons/attributes/stringtable.xml
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project name="ZEN">
     <Package name="Attributes">
+        <Key ID="STR_ZEN_Attributes_Medic">
+            <English>Medic</English>
+        </Key>
```
dedmen: From languagecore_f

```
<Key ID="str_support_medic">
  <Original>Medic</Original>
  <French>Infirmier</French>
  <Spanish>Médico</Spanish>
  <Italian>Medico</Italian>
  <Polish>Medyk</Polish>
  <Russian>Санинструктор</Russian>
  <German>Sanitäter</German>
  <Czech>Zdravotník</Czech>
  <English>Medic</English>
  <Chinese>醫護兵</Chinese>
  <Portuguese>Médico</Portuguese>
  <Korean>의무병</Korean>
  <Chinesesimp>军医</Chinesesimp>
  <Turkish>Sıhhiye</Turkish>
  <Japanese>医者</Japanese>
</Key>
```

CreepPork:
I'm starting to think, you have a separate entry for the medic attribute. But what then happens if you want to add an engineer slot and EOD (from ACE)? Are you going to create 2 more rows? I think that will take up too much space and make it cluttered.

mharis001:
I don't think there is a way to avoid 3 rows of attributes for: medic, engineer, and EOD. Maybe checkboxes instead of toolboxes would look better.

CreepPork:
What about a button like `Inventory` for the skills?

jonpas:
Mind, not all are Yes/No, but have 3 or more step setting.

mharis001:
Yeah, will need to add special handling for ACE. I'm not sure on putting this under a "buttons display". It would make the main attributes display cleaner but add an extra step in accessing this attribute.

CreepPork:
It's not like you are going to be selecting an entire squad of units to set as medics. It doesn't have to get accessed very often. You could append this to #42.

PR #46, opened 2019-07-02
Fix attach flag dropdown list

### LINKED ISSUE #21: Bug: Attach flag dropdown list not clickable
**Arma 3 Version:** `1.92`
**CBA Version:** `3.11.2`
**ZEN Version:** `1.1.0`

**Mods:**
```
Zeus Enhanced 1.1.0
[SOB] BW EinsUstg v3.4.0
[SOB] SOBMod v5.0.0
Community Base Addons v3.11.2
Bundeswehr Mod Compatibility PBO v2.1-1
Bundeswehr Mod
Advanced Combat Radio Environment 2
Advanced Combat Environment Extras 3.4.2
Advanced Combat Environment 3.12.6
```

**Description:**
- The options in the "Attach Flag" dialog are not clickable. The only working navigation for me is using the arrow keys on the keyboard

**Steps to reproduce:**
- Open Zeus
- Activate the Attach Flag module
- Click on a unit with it
- Try to select a flag by clicking on it

- Dedicated

**Additional information:**

b19afcca3375eb7d5d61d5b494ea77da

mharis001:
- Fix #21 by changing flags list from a combo box to a list box
- Add list box control type to dialog
- `gui_list` function shares some code with `gui_combo` but I don't think there is a clear way to avoid that


addons/dialog/functions/fnc_create.sqf
```diff
@@ -67,6 +67,42 @@ scopeName "Main";
...
+        _labels resize count _values;
+
+        _labels = _labels apply {
+            if (isNil "_x") then {
+                _x = _values select _forEachIndex;
+            };
+
+            _x params ["_label", ["_tooltip", ""], ["_picture", "", [""]], ["_textColor", [1, 1, 1, 1], [[]], 4]];
+
+            if !(_label isEqualType "") then {
```
CreepPork: Couldn't you move this to `params` so it yells if anything has gone wrong with the data types? Yes, it wouldn't then continue but it's better than a silent error which then processes your data.

mharis001: I think this is here for the `_x = _values select _forEachIndex` above but I agree, it would be better to just move this up to the `params`.

PR #47, opened 2019-07-03
Object placement preview

synixebrett:
- Display a preview of vehicles while placing
- Use `lineIntersectsSurfaces` to place objects


addons/placement/functions/fnc_updatePreview.sqf
```diff
@@ -0,0 +1,62 @@
...
+	GVAR(previewVehicle) = objNull;
+};
+
+// Create the preview if it doesn't exist
+if (isNull GVAR(previewVehicle)) then {
+    GVAR(previewObject) = "Land_Wrench_F" createVehicleLocal (ASLtoAGL _pos);
+	GVAR(previewObject) hideObject true;
+	GVAR(previewVehicle) = _class createVehicleLocal [0, 0, 0];
+	GVAR(previewVehicle) attachTo [GVAR(previewObject), [0, 0, OFFSET + 1 + ((boundingCenter GVAR(previewVehicle)) select 2)]];
+	GVAR(previewVehicle) enableSimulationGlobal false;
```
jonpas: Use createSimpleObject with local argument instead (next update, another PR)?

addons/placement/functions/fnc_updatePreview.sqf
```diff
@@ -0,0 +1,62 @@
...
+		deleteVehicle GVAR(previewObject);
+		GVAR(previewVehicle) = objNull;
+	};
+};
+
+// Delete the old preview if the player selects a different item
+if !(_class isEqualTo (typeOf GVAR(previewVehicle))) then {
+	deleteVehicle GVAR(previewVehicle);
+	deleteVehicle GVAR(previewObject);
+	GVAR(previewVehicle) = objNull;
```
jonpas: Indentation trouble. Ia there no EditorConfig in this repo yet? Someone should add it!

addons/placement/functions/fnc_updatePreview.sqf
synixebrett: This is the plan once that hits the main branch

addons/placement/functions/fnc_updatePreview.sqf
CreepPork: Editor config is present but I guess their editors don't support it or it's disabled.

addons/placement/functions/fnc_getSelectedClass.sqf
```diff
@@ -0,0 +1,4 @@
+{
+  private _faction = (findDisplay 312) displayCtrl _x;
+  if (ctrlShown _faction) exitWith { _faction tvData (tvCurSel _faction) };
+} forEach [270, 271, 272, 273, 274];
```
CreepPork: https://ace3mod.com/wiki/development/coding-guidelines.html#56-magic-numbers

addons/placement/functions/fnc_boundingBoxAdjustment.sqf
```diff
@@ -0,0 +1,55 @@
...
+			} count _bbx;
+		} count _bbz;
+		reverse _bbz;
+	} count _bby;
+	_arr pushBack (_arr select 0);
+	_arr pushBack (_arr select 1);
+	_arr
+};
+bbox = boundingBox _obj call _bb;
+bboxr = boundingBoxReal _obj call _bb;
```
CreepPork: Why are these even globals? And [they should be prefixed](https://ace3mod.com/wiki/development/coding-guidelines.html#111-global-variable-naming)

addons/placement/XEH_postInit.sqf
```diff
@@ -0,0 +1,29 @@
...
+
+["ZEN_displayCuratorLoad", {
+    params ["_display"];
+    GVAR(previewDraw) = addMissionEventHandler ["Draw3D", {
+        call FUNC(updatePreview);
+    }];
+}] call CBA_fnc_addEventHandler;
+
+
+[{!isNull (getAssignedCuratorLogic player)}, {
```
veteran29: Wouldn't it be better to just add this EH to all curators on their init? Curators EH only fire when curator logic is local and it will be local to the player that controls it AFAIK.

This WUAE will stay there forever for most of the players eating that micro seconds of execution time :P

addons/placement/functions/fnc_updatePreview.sqf
```diff
@@ -0,0 +1,62 @@
...
+};
+
+// Create the preview if it doesn't exist
+if (isNull GVAR(previewVehicle)) then {
+    GVAR(previewObject) = "Land_Wrench_F" createVehicleLocal (ASLtoAGL _pos);
+    GVAR(previewObject) hideObject true;
+    GVAR(previewVehicle) = _class createVehicleLocal [0, 0, 0];
+    GVAR(previewVehicle) attachTo [GVAR(previewObject), [0, 0, OFFSET + 1 + ((boundingCenter GVAR(previewVehicle)) select 2)]];
+    GVAR(previewVehicle) enableSimulationGlobal false;
+    GVAR(previewVehicle) disableCollisionWith player;
```
CreepPork: Wouldn't this only disable the collision with the Zeus?

synixebrett: Yes, I only had an issue with it colliding with the local player

addons/placement/functions/fnc_boundingBoxAdjustment.sqf
synixebrett: They are not globals, this file is being removed anyway before this PR is marked ready. It was being used for development

CreepPork: Alright, then nevermind.

addons/placement/functions/fnc_getSelectedClass.sqf
```diff
@@ -0,0 +1,6 @@
+#include "script_component.hpp"
```
CreepPork: Doc block

addons/placement/functions/fnc_updatePreview.sqf
```diff
@@ -0,0 +1,59 @@
+#include "script_component.hpp"
```
CreepPork: Doc block

CreepPork:
Same as #49, shouldn't this go under the `editor` component as it's directly related how the Zeus editor works?

addons/placement/functions/fnc_updatePreview.sqf
dedmen: It will be in Contact DLC update version 1.94 on the 27th.

mharis001:
With `createSimpleObject` (local), I think this will have to be changed to work with the empty (prop) objects only since, simple infantry unit objects have no heads and are in a t-pose.

Also, the preview object's position is slightly above (1m?) the surface position. Any specific reason for that?

synixebrett:
It should need to do all of them.

The one meter is because of the placements changes to not collide with surfaces, like aircraft carriers

synixebrett:
This needs a lot of work anyway and may not be viable in the end, it would be cool but not something game changing

mharis001:
I switched this to an events based approach instead of a constantly running Draw3D for determining when to create/delete the preview. Also, added a setting to control if this is enabled.

Switched to using the difference between the world position and land contact transformed position to determine the vertical attach offset. Produces better results (pretty much perfect for the objects I tested) than just using `boundingCenter` - which placed some objects into the ground.

Currently doesn't include the shift key to rotate preview, because it was a bit rough - can implement if a better approach is found, but also might not be necessary.

Some objects may still get destroyed if you place them in weird positions, however, I think that is just going to be up to the user to use wisely. From my testing, placing on relatively flat surfaces works well.

I think this should be good to go @synixebrett .

synixebrett:
I'll give it a run this Monday

addons/placement/functions/fnc_setupPreview.sqf
```diff
@@ -0,0 +1,53 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * ["B_Soldier_F"] call zen_placement_fnc_setupPreview
+ *
+ * Public: No
+ */
+
+params ["_objectType"];
```
CreepPork: This could take by default an empty string, therefore, cleaning up code some more.

PR #49, opened 2019-07-03
Player visibility

synixebrett:
- Indicates to Zeus when a player can see their cursor

Been using this over a year now and feel weird when it isn't there, prevents Zeus from placing objects in the view of players

image
Imagine a cursor above VISIBLE at the end of the red line


addons/visibility/config.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+class CfgPatches {
+    class ADDON {
+        name = COMPONENT_NAME;
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"zen_common"};
+        author = ECSTRING(main,Author);
+        authors[] = {"Brett"};
```
CreepPork: #47 had `SynxieBrett`

CreepPork:
I'm not so certain but shouldn't this go under the `editor` component?

addons/visibility/config.cpp
synixebrett: I know, I'm going to change #47 to match the AUTHORS.txt

CreepPork:
Also, maybe not red but white (customizable?) and perhaps an icon as well? The white color looks fine for the vanilla laser designator line.

mharis001:
I think it's good in its own component. No harm in separating bigger features out into individual components.

jonpas:
Red signifies warning/danger that they can see it, it's fine IMO.

CreepPork:
Yeah, you're right, but still could use some customizability. And I'm second guessing the icon, as the screenshot doesn't have the cursor. But maybe an icon (a gear?) with the same color could be drawn behind the cursor to more illustrate the _danger_.

addons/visibility/XEH_postInit.sqf
```diff
@@ -0,0 +1,49 @@
...
+        if ((count _intersections) != 0) then {
+            _pos = ((_intersections select 0) select 0);
+        };
+
+        // check 1.5 above the cursor to prevent a small object on the terrain blocking the view
+        private _posHigh = [_pos select 0, _pos select 1, (_pos select 2) + 1.5];
+        private _draw = false;
+        {
+            if (side _x != sideLogic) then {
+                private _dir = (_x getRelDir _posHigh) + 90;
```
diwako: Suggesting this instead of the if check
```sqf
 private _dir = ((_x getRelDir _posHigh) + 90) mod 360;
```

addons/visibility/XEH_postInit.sqf
```diff
@@ -0,0 +1,49 @@
...
+        } forEach allPlayers;
+
+        // Write visibile under the cursor if any player can see the position
+        if (_draw) then {
+            drawIcon3D [
+                "",
+                [1, 0, 0, 1],
+                ASLToAGL _pos,
+                1, 3, 0,
+                "VISIBLE",
```
diwako: Possibility for string table support?

addons/visibility/XEH_postInit.sqf
```diff
@@ -0,0 +1,49 @@
...
+    GVAR(visibilityDraw) = addMissionEventHandler ["Draw3D", {
+        // The the cursor position in the world
+        private _pos = AGLtoASL screenToWorld getMousePosition;
+        private _intersections = lineIntersectsSurfaces [getPosASL curatorCamera, _pos];
+        if ((count _intersections) != 0) then {
+            _pos = ((_intersections select 0) select 0);
+        };
+
+        // check 1.5 above the cursor to prevent a small object on the terrain blocking the view
+        private _posHigh = [_pos select 0, _pos select 1, (_pos select 2) + 1.5];
```
diwako: Not sure if this is faster, it seems to me more readable.
```sqf
private _posHigh = _pos vectorAdd [0, 0, 1.5];
```

dedmen:
This needs to be disable-able (performance cost). CBA setting?

addons/visibility/XEH_postInit.sqf
```diff
@@ -0,0 +1,49 @@
...
+        private _draw = false;
+        {
+            if (side _x != sideLogic) then {
+                private _dir = (_x getRelDir _posHigh) + 90;
+                if (_dir >= 360) then {
+                    _dir = _dir - 360;
+                };
+                if (_dir < 180) then {
+                    if (count lineIntersectsSurfaces [eyePos _x, _pos, _x, objNull] == 0 || {count lineIntersectsSurfaces [eyePos _x, _posHigh, _x, objNull] == 0}) then {
```
dedmen: count array == 0 -> `array isEqualTo []`

addons/visibility/XEH_postInit.sqf
```diff
@@ -0,0 +1,49 @@
...
+                        // Check visibility through smoke
+                        private _visibility = [objNull, "VIEW"] checkVisibility [eyePos _x, _posHigh];
+                        // Draw a line from each player that can see the cursor
+                        drawLine3D [ASLToAGL eyePos _x, ASLToAGL _pos, [1, 0, 0, _visibility]];
```
dedmen: consider not retrieving eyePos 3 times. Use a variable

diwako:
I played around with it a bit, noticed the EH will keep running when zeus is closed and will be added again once zeus display is opened again. Meaning you will just layer this Draw3d logic over and over.

So I suggest adding an eh when zeus display is closed.
```sqf
GVAR(visibilityDraw) = -1;
["featureCamera", {
    params ["", "_featureCamera"];
    if (GVAR(visibilityDraw) >= 0 && {_featureCamera isEqualTo ""}) then {
		removeMissionEventHandler ["Draw3D", GVAR(visibilityDraw)];
		GVAR(visibilityDraw) = -1;
	};
}, true] call CBA_fnc_addPlayerEventHandler;
```

I tried using the `ZEN_displayCuratorUnload` event, but that seems to just fire directly after the `ZEN_displayCuratorLoad` event.

synixebrett:
My plan was to add the CuratorUnload event. If that is not working as expected that should be fixed first, and then have this PR use it. Can you create an issue for `ZEN_displayCuratorUnload`

mharis001:
Just tested the `ZEN_displayCuratorLoad` and `ZEN_displayCuratorUnload` events and they are working as expected for me. @diwako can you provide your usage/details.

addons/visibility/functions/fnc_start.sqf
```diff
@@ -0,0 +1,55 @@
+#include "script_component.hpp"
+/*
```
mharis001: Use full function header (arguments/return).

addons/visibility/functions/fnc_start.sqf
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * Example:
+ * [] call zen_visibility_fnc_start
+ *
+ * Public: No
+ */
+
+if (GVAR(draw) != -1) exitWith {};
+
+GVAR(draw) = addMissionEventHandler ["Draw3D", {
```
mharis001: This needs to be put into a separate function. [reference](https://ace3mod.com/wiki/development/coding-guidelines.html#73-event-driven)

addons/visibility/script_component.hpp
```diff
@@ -0,0 +1,18 @@
+#define COMPONENT visibility
+#define COMPONENT_BEAUTIFIED Visibility
+#include "\x\zen\addons\main\script_mod.hpp"
+
+// #define DEBUG_MODE_FULL
+// #define DISABLE_COMPILE_CACHE
```
mharis001: Should add `// #define ENABLE_PERFORMANCE_COUNTERS`.

addons/visibility/stringtable.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project name="ZEN">
+    <Package name="Visibility">
+        <Key ID="STR_ZEN_Visibility_Enabled_Description">
+            <English>Indicates when a player can see the Zeus' cursor</English>
```
mharis001: Needs a period at the end, for consistency with other settings.

Maybe "cursor position" instead of just "cursor", but i am not sure on that.

addons/visibility/functions/fnc_stop.sqf
```diff
@@ -0,0 +1,15 @@
+#include "script_component.hpp"
+/*
```
mharis001: Use full function header.

addons/visibility/functions/fnc_start.sqf
```diff
@@ -0,0 +1,55 @@
...
+    // The the cursor position in the world
+    private _pos = AGLtoASL screenToWorld getMousePosition;
+    private _intersections = lineIntersectsSurfaces [getPosASL curatorCamera, _pos];
+    if ((count _intersections) != 0) then {
```
mharis001: Use `isEqualTo []` instead.

addons/visibility/functions/fnc_start.sqf
```diff
@@ -0,0 +1,55 @@
...
+        _pos = ((_intersections select 0) select 0);
+    };
+
+    // check 1.5 above the cursor to prevent a small object on the terrain blocking the view
+    private _posHigh = _pos vectorAdd [0, 0, 1.5];
+    private _draw = false;
+    {
+        if (side _x != sideLogic) then {
+            private _dir = ((_x getRelDir _posHigh) + 90) mod 360;
```
mharis001: `_dir` variable only used in the if condition right below. Can also put both checks in the if statement above?

addons/visibility/XEH_postInit.sqf
```diff
@@ -0,0 +1,23 @@
...
+}] call CBA_fnc_addEventHandler;
+
+["ZEN_displayCuratorUnload", {
+    call FUNC(stop);
+}] call CBA_fnc_addEventHandler;
+
+["CBA_SettingChanged", {
+    params ["_setting", "_value"];
+    if (_setting isEqualTo QGVAR(enabled)) then {
+        if (isNull (findDisplay IDD_RSCDISPLAYCURATOR)) exitWith {};
```
mharis001: I think this can be moved to the top.

synixebrett: Wouldn't the setting name check be faster?

dedmen: you could make both of them an exitWith.
I  think that would look nicer.

synixebrett: Ah right, forgot that existed. I'll switch to that

addons/visibility/functions/fnc_draw.sqf
```diff
@@ -0,0 +1,54 @@
+#include "script_component.hpp"
+/*
+ * Author: Brett
+ * Checks and draws the visibility indication
```
mharis001:
```suggestion
 * Checks and draws the visibility indicator.
```

addons/visibility/functions/fnc_draw.sqf
```diff
@@ -0,0 +1,54 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call zen_visibility_fnc_draw
+ *
+ * Public: No
+ */
+
+// The the cursor position in the world
```
mharis001:
```suggestion
// The cursor position in the world
```

addons/visibility/functions/fnc_start.sqf
```diff
@@ -0,0 +1,20 @@
+#include "script_component.hpp"
+/*
+ * Author: Brett
+ * Starts the Draw3D eh that indicates when a player can see the curator cursor
```
mharis001:
```suggestion
 * Starts the Draw3D eh that indicates when a player can see the curator cursor.
```

addons/visibility/functions/fnc_stop.sqf
```diff
@@ -0,0 +1,21 @@
...
+ * Author: Brett
+ * Removes the Draw3D event handler
```
mharis001:
```suggestion
 * Removes the Draw3D event handler.
```

CreepPork:
I think it would be for the better if it would be disabled by default because it would add a performance cost and I don't think all the users would want to see it.

jonpas:
Disablable but enabled by default sounds better. One line is not going to cause a noticeable performance difference.

dedmen:
>One line is not going to cause a noticeable performance difference.

One line, per nearby player, per frame.

synixebrett:
I've used this with up to 21 people, never had the slightest issue

mharis001:
I think this is good to go. Can always change the default later if needed.

PR #51, opened 2019-07-03
Added Azure Pipelines support for CI builds

CreepPork:
- Add a makefile for building PBOs
- Add Azure Pipelines support
- Add `getArmake.ps1` PowerShell script for Windows users to download armake
~~Play around and test with Travis~~
- [x] Add notifications


jonpas:
armake (1) is no longer supported!

Use HEMTT and you don't need a long Makefile or a PowerShell download script.

CreepPork:
But armake is more stable than armake2 and doesn't require any more changes to the code base than proposed here.

dedmen:
Then you might aswell look at armake++ which is faster and has even more bugs fixed.
https://github.com/dedmen/armake/releases
not sure how up-to-date the release files are compared tothe code.

CreepPork:
I'm not sure I should migrate over to armake++, it's plenty fast enough, the build time is less than 1 minute and for now, no bugs have appeared, so it's pretty stable.

addons/main/script_version.hpp
```diff
@@ -1,4 +1,4 @@
 #define MAJOR 1
 #define MINOR 1
 #define PATCHLVL 0
-#define BUILD 4
+#define BUILD 4
```
mharis001: Newline?

CreepPork: The new line is stripped by the makefile, I’ll see if I can append `\n`.

jonpas: You are appending a newline at the others, why not the last one.

PR #52, opened 2019-07-03
Add healing context action

3Mydlo3:
Heals all units in given selection. Works for units in vehicles.

- Add healing context action
- Allow healing of all units in selection, only AIs or only players.

_Icon copied from ACE medical._


addons/context_actions/functions/fnc_healPlayers.sqf
```diff
@@ -0,0 +1,31 @@
...
+ * Public: No
+ */
+#include "script_component.hpp"
+
+private _players = _this select {isPlayer _x};
+
+private _moduleGroup = createGroup sideLogic;
+{
+	private _module = _moduleGroup createUnit [
+		QEGVAR(modules,moduleHeal),
```
jonpas: This will throw a bunch of error messages for all non-`CAManBase` units.

A better way to do this would be to move the healing logic to some common function, then calling that directly from here.

3Mydlo3: I thought about healing logic transfer, but didn't want to do it unless told to. I'll make common function variant then.

addons/common/functions/fnc_healUnit.sqf
```diff
@@ -0,0 +1,30 @@
...
+    ["ace_medical_treatmentAdvanced_fullHealLocal", [_unit, _unit], _unit] call CBA_fnc_targetEvent;
+} else {
+    // BI's scripted revive system
+    if ((missionNamespace getVariable ["bis_revive_mode", 0]) != 0) then {
+        ["#rev", 1, _unit] call BIS_fnc_reviveOnState;
+        _unit setVariable ["#rev", 1, true];
+    } else {
+        _unit setDamage 0;
+    };
+};
```
mharis001: Newline.

addons/common/functions/fnc_healUnit.sqf
```diff
@@ -0,0 +1,30 @@
+/*
+ * Author: mharis001, 3Mydlo3
+ * Heals unit
```
mharis001: Maybe: `Heals the given unit.`

addons/context_actions/functions/fnc_healPlayers.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * None
+ *
+ * Example:
+ * [player1, player2, player3] call zen_context_actions_fnc_healPlayers
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+private _mans = [];
```
mharis001: Maybe `_units` is a better variable name 😄.

addons/context_actions/functions/fnc_healPlayers.sqf
```diff
@@ -0,0 +1,25 @@
...
+{
+	_mans append crew _x;
+} forEach _this;
+
+{
+	[_x] call EFUNC(common,healUnit);
+} forEach _mans select {isPlayer _x};
```
mharis001: I think it is better to get rid of the select here and move the `isPlayer` check inside the `forEach` loop. 2 loops vs 1.

addons/modules/functions/fnc_moduleHeal.sqf
```diff
@@ -32,14 +32,4 @@ if (!alive _unit) exitWith {
...
-} else {
-    // BI's scripted revive system
-    if ((missionNamespace getVariable ["bis_revive_mode", 0]) != 0) then {
-        ["#rev", 1, _unit] call BIS_fnc_reviveOnState;
-        _unit setVariable ["#rev", 1, true];
-    } else {
-        _unit setDamage 0;
-    };
-};
+[_unit] call EFUNC(common,healUnit);
```
mharis001: Newline.

CreepPork:
@3Mydlo3 Make sure your editor supports `.editorconfig` files.

3Mydlo3:
@CreepPork Whoops, I didn't realize I need particular extension for editorconfig in VS Code. I'll fix everything ASAP.

addons/context_actions/functions/fnc_healPlayers.sqf
```diff
@@ -0,0 +1,27 @@
...
+ * Example:
+ * [player1, player2, player3] call zen_context_actions_fnc_healPlayers
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+private _units = [];
+{
+    _units append crew _x;
```
CreepPork: Shouldn't this be using `fullCrew`? As in if you use it on a vehicle, currently, you would only heal the - driver, gunner, commander. But what about the passengers in the vehicle?

And for that matter, if you select a single unit (from the unit tree) and open the context menu on him and press heal, then it will heal everyone in the vehicle, even though you wanted only 1 unit to get healed.

3Mydlo3: It won't heal passengers, that's true, I'll fix this.

But I don't agree on the second part. If I select 1 guy in a vehicle (using unit tree), `curatorSelected` returns only this selected unit, and later on `crew _x` also returns this one unit. Maybe I cannot see something obvious, feel free to correct me.

CreepPork: 🤔 I’ll take a look later in game. Maybe I’m confusing something.

3Mydlo3: I performed further tests and now I don't agree with first part either. `crew` according to wiki returns "crew in order [driver,gunner,commander,turrets,cargo]" and my tests confirm that, so everyone on board would be healed (as long as it's player).

There might be just one issue, when we select vehicle and it's crew and then click heal, as everyone will be healed twice, but it can be easily fixed by adding `arrayIntersect`.

CreepPork: Yeah, you're right, it looks like I was jumping to conclusions from the first line of the BIKI page :smile: 
And yeah, that should be addressed.

CreepPork:
And overall, is there a specific reason this is only limited to players? Seems like AI could benefit from this as well.

3Mydlo3:
Yes, when AIs are selected you cannot open context menu, you just set them waypoints. I thought like you yesterday and prepared everything only to discover this issue when everything was ready.

CreepPork:
Is it together with player units (Player + AI)? Or is it just AI? I thought I could open the context menu for AI.

3Mydlo3:
I tried selecting players and AIs and ended up placing waypoints. It works for them unless they are selected (so only on hover) and I didn't want to make it too confusing that you can heal multiple players but only one AI at a time and everything under one option. It just doesn't seem good to me.

CreepPork:
Perhaps this is a bug? Seems like really weird behaviour to me. @mharis001

3Mydlo3:
Ok, I checked in game and confirmed (I wanted to be 100% sure that I didn't miss anything) that when AI leads a group RMB places a waypoint, doesn't open context menu.

3Mydlo3:
Ok, I didn't notice that we have separate keybind for context menu, it's `V` by default, i used RMB all the time. Sorry for misinformation.

3Mydlo3:
I will add healing for AIs then, just like formations or behaviour are done. It would be beneficial indeed.

addons/common/functions/fnc_healUnit.sqf
```diff
@@ -0,0 +1,30 @@
...
+ * [player] call zen_common_fnc_healUnit
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_unit"];
+
+if (GVAR(aceMedicalLoaded) && {ace_medical_level > 0}) then {
+    ["ace_medical_treatmentAdvanced_fullHealLocal", [_unit, _unit], _unit] call CBA_fnc_targetEvent;
```
dedmen: This will change on next ACE update to
`ace_medical_treatment_fnc_fullHealLocal`
You can detect the next ace update with CfgPatches class "ACE_medical_treatment"
https://github.com/acemod/ACE3/tree/master/addons/medical_treatment

3Mydlo3: I will add a check then, thanks for the info. 👍

addons/context_actions/functions/fnc_healUnits.sqf
```diff
@@ -0,0 +1,36 @@
...
+ * Author: 3Mydlo3, veteran29
+ * Heals units in given selection. Works for units in vehicles.
+ * Healing modes:
+ * HEAL_MODE_ALL - 0
+ * HEAL_MODE_AI - 1
+ * HEAL_MODE_PLAYERS - 2
+ *
+ * Arguments:
+ * 0: Selected Objects <ARRAY>
+ * 1: Mode <SCALAR>
```
mharis001: Rest of the project uses `<NUMBER>`.

addons/context_actions/functions/fnc_healUnits.sqf
```diff
@@ -0,0 +1,36 @@
...
+#include "script_component.hpp"
+
+params ["_objects", "_mode"];
+
+private _units = [];
+{
+    _units append crew _x;
+} forEach _objects;
+
+private _fnc_filter = [{true}, {!isPlayer _this}, {isPlayer _this}] select _mode;
```
mharis001: Can replace the `_this` with `_x` and then in the if condition just do `alive _x && _fnc_filter`.

addons/context_actions/CfgContext.hpp
```diff
@@ -199,6 +199,29 @@ class EGVAR(context_menu,actions) {
...
+            statement = QUOTE([ARR_2(_selectedObjects, HEAL_MODE_AI)] call FUNC(healUnits));
+            priority = 2;
+        };
+        class Players {
+            displayName = ECSTRING(modules,Players);
+            condition = QUOTE(_selectedObjects findIf {crew _x findIf {isPlayer _x && {alive _x}} != -1} != -1);
+            statement = QUOTE([ARR_2(_selectedObjects, HEAL_MODE_PLAYERS)] call FUNC(healUnits));
+            priority = 1;
+        };
+        priority = -98;
```
mharis001: I think this action might be better suited higher up in the menu, under stance maybe: `priority = -70`?

addons/context_actions/CfgContext.hpp
```diff
@@ -199,6 +199,29 @@ class EGVAR(context_menu,actions) {
         statement = QUOTE(_hoveredEntity call EFUNC(garage,openGarage));
         priority = -82;
     };
+    class HealUnits {
+        displayName = ECSTRING(modules,ModuleHeal);
+        icon = QPATHTOF(ui\medical_cross_ca.paa);
+        condition = QUOTE(_selectedObjects findIf {crew _x findIf {alive _x} != -1} != -1);
+        class All {
+            displayName = ECSTRING(common,All);
+            statement = QUOTE([ARR_2(_selectedObjects, HEAL_MODE_ALL)] call FUNC(healUnits));
```
mharis001: No space after comma in the `ARR_2`.

addons/context_actions/CfgContext.hpp
```diff
@@ -199,6 +199,29 @@ class EGVAR(context_menu,actions) {
         statement = QUOTE(_hoveredEntity call EFUNC(garage,openGarage));
         priority = -82;
     };
+    class HealUnits {
+        displayName = ECSTRING(modules,ModuleHeal);
+        icon = QPATHTOF(ui\medical_cross_ca.paa);
+        condition = QUOTE(_selectedObjects findIf {crew _x findIf {alive _x} != -1} != -1);
```
mharis001: The action(s) shows when a module is in the selection. I think this needs a `_x isKindOf 'CAManBase'` check.

addons/context_actions/functions/fnc_healUnits.sqf
```diff
@@ -0,0 +1,36 @@
...
+
+{
+    if (alive _x && {_x call _fnc_filter}) then {
```
mharis001: Similar to the conditions, might be good to check for `CAManBase` here.

mharis001:
@CreepPork The context menu not opening when waypoint placement is possible is not a bug. 
>Since it is not possible to overwrite the "place waypoint on right-click" behavior when AI units are selected in Zeus, the context menu can be opened using both the right mouse button and CBA keybind (default: V).

PR #55, opened 2019-07-03
Add HEMTT support

CreepPork:
- Add HEMTT support with download script
- [x] Currently limited by armake2 as it throws errors for dik codes
- [x] As far as I can see, I can't disable specific armake2 warnings in the HEMTT config file
- [x] Blocked by synixebrett/HEMTT#267 (for Linux users)
- [ ] With #245, HEMTT now displays a warning about compats:
```
warning: compatibility addon `compat_ace` should be in `compats/`
```


dedmen:
Guess same as ACE/CBA this will have to wait till the armake2 preprocessor issues are fixed.

hemtt.json
```diff
@@ -0,0 +1,25 @@
...
+        "./include"
+    ],
+    "headerexts": [
+        "version={{git \"id 8\"}}"
+    ],
+    "sigversion": 3,
+    "keyname": "{{prefix}}_{{version}}",
+    "signame": "{{prefix}}_{{version}}-{{git \"id 8\"}}",
+    "releasebuild": [
+        "@zip ace3_{{semver.major}}.{{semver.minor}}.{{semver.patch}}"
```
CreepPork:
```suggestion
        "@zip zen_{{semver.major}}.{{semver.minor}}.{{semver.patch}}"
```

bux:
Make sure that keyname and signame are equal.
There's currently an issue with servers failing to match different named keys/sigs.

.gitignore
```diff
@@ -12,5 +12,7 @@ CHANGELOG.md
 *.bisign
 *.tar.gz
 *.zip
+hemtt*
```
CreepPork: Just use ./hemtt

hemtt.toml
```diff
@@ -0,0 +1,26 @@
+name = "Zeus Enhanced"
+prefix = "zen"
+author = "ZEN Team"
+mainprefix = "x"
+sigversion = 3
+version = "1.6.1"
+reuse_private_key = false
```
CreepPork: I think this is by default false

hemtt.toml
```diff
@@ -0,0 +1,26 @@
+name = "Zeus Enhanced"
+prefix = "zen"
+author = "ZEN Team"
+mainprefix = "x"
+sigversion = 3
+version = "1.6.1"
```
CreepPork: Don’t use this, makes version numbers a pain because you’d have to update this one too.

It already looks for script_version

tools/config_style_checker.py
```diff
@@ -0,0 +1,157 @@
...
+import re
+import ntpath
+import sys
+import argparse
+
+def check_config_style(filepath):
+    bad_count_file = 0
+    def pushClosing(t):
+        closingStack.append(closing.expr)
+        closing << Literal( closingFor[t[0]] )
```
CreepPork: Remove double spaces

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Validate Config
+      run: python3 tools/config_style_checker.py
+    - name: Check for BOM
+      uses: arma-actions/bom-check@master
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout the source code
+      uses: actions/checkout@master
+    - name: Lint (sqflint)
```
CreepPork: Looking back from Achilles, we can remove this to save build times and massive logs because this isn’t really an improvement due to false positives.

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - name: Build addon with HEMTT
+        uses: arma-actions/hemtt@master
+        with:
+          command: 'build --release'
+      - uses: actions/upload-artifact@master
+        with:
+          name: '@ZEN'
```
CreepPork: I don’t know but this will already prefix it but it does need the version number.

jonpas: Upload artifact is really rudimentary, you can't even properly upload a zip without double-zipping it. Attaching a proper version is its own issue. Environment variables maybe? Ping me if you make if work! 

I commented it out in a few of my projects.

.github/workflows/build.yml
neilzar: I'll remove it, I added it just to test this out.

.gitignore
neilzar: You were the one to add this, but I'll change it.

CreepPork: Whoops, then I guess my mind changed :D

@jonpas does the template folder need to be gitignored or is it safe to include in git?

.github/workflows/build.yml
neilzar: @jonpas V2 of upload-artifact will support Glob wildcards, haven't figured out how to not have it double-zip though.

CreepPork:
The pipelines yaml can be removed anyway if we're going all out with Actions.

.gitignore
jonpas: It's your template, entirely depends where you want to host it. However, considering it's not even fully implemented in current HEMTT, it doesn't matter either way.

.github/workflows/build.yml
jonpas: Yeah, I am still waiting for that.

.gitignore
CreepPork: Then I guess, for now, we can gitignore it.

hemtt.toml
```diff
@@ -0,0 +1,24 @@
+name = "Zeus Enhanced"
+prefix = "zen"
+author = "ZEN Team"
+mainprefix = "x"
+sigversion = 3
```
synixebrett: This is default and could also be removed

hemtt.toml
```diff
@@ -0,0 +1,24 @@
...
+    "mod.cpp",
+    "logo_zen_ca.paa",
+    "logo_zen_small_ca.paa",
+    "LICENSE",
+    "AUTHORS.txt",
+    "README.md"
+]
+include = ["./include"]
+
+keyname = "{{prefix}}_{{version}}"
```
synixebrett:
```suggestion
key_name = "{{prefix}}_{{version}}"
```

hemtt.toml
```diff
@@ -0,0 +1,24 @@
...
+signame = "{{prefix}}_{{version}}-{{git \"id 8\"}}"
```
synixebrett:
```suggestion
authority = "{{prefix}}_{{version}}-{{git \"id 8\"}}"
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
+name: Validate and Build
```
CreepPork: I'd prefer something cleaner (one word) because currently, the messages aren't as clear as they could be:
```
Validate and Build / validate (pull_request) Successful in 14s
Validate and Build / build (pull_request) Successful in 25s
```

It would be more understandable if it was something like:
```
Compile / validate (pull_request) Successful in 14s
Compile / build (pull_request) Successful in 25s
```

Note, this is not final, I'm just proposing something shorter (build, compile, that sort of names).
```suggestion
name: Compile
```

.gitignore
```diff
@@ -12,5 +13,7 @@ CHANGELOG.md
...
+!hemtt.toml
```
CreepPork:
```suggestion
./.hemtt
```
This should gitignore the template folder but not the `toml` file.

hemtt.toml
```diff
@@ -0,0 +1,23 @@
...
+    "AUTHORS.txt",
+    "README.md"
+]
+include = ["./include"]
+
+key_name = "{{prefix}}_{{version}}"
+authority = "{{prefix}}_{{version}}-{{git \"id 8\"}}"
+
+releasebuild = [
+    "@zip @ZEN-{{version}}"
```
mharis001:
```suggestion
    "@zip @zen_{{version}}"
```

hemtt.toml
```diff
@@ -0,0 +1,23 @@
...
+]
+
+[header_exts]
+version= "{{git \"id 8\"}}"
```
mharis001:
```suggestion
version = "{{git \"id 8\"}}"
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
```
mharis001: All branches?

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: Check for BOM
+      uses: arma-actions/bom-check@master
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the source code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - name: Build addon with HEMTT
```
mharis001:
```suggestion
      - name: Build with HEMTT
```

.gitignore
```diff
@@ -1,4 +1,5 @@
-release/*
```
mharis001: Still need this for `make.py`.

hemtt.toml
```diff
@@ -0,0 +1,23 @@
+name = "Zeus Enhanced"
+prefix = "zen"
+author = "ZEN Team"
+mainprefix = "x"
+files = [
```
mharis001: Also needs `meta.cpp`.

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
+name: Build
```
mharis001: Maybe `Arma`?

.github/workflows/build.yml
CreepPork: It will trigger under these conditions:
- If the branch is a pull request which targets `master`
- Pushed directly to `master`.

I think it's pretty fine as is.

.github/workflows/build.yml
CreepPork: Don't think this conveys the message at all.

.github/workflows/build.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Build with HEMTT
         uses: arma-actions/hemtt@master
         with:
-          command: 'build --release'
+          command: build --release --ci
```
synixebrett: Was it not detecting CI support

neilzar: Seems it wasn't. It started doing the checks and prebuild after adding this.

.github/workflows/build.yml
```diff
@@ -33,25 +38,12 @@ jobs:
...
-      #     path: releases/@zen-*.zip
-  notify:
-    needs: [validate, build]
-    if: always()
-    runs-on: ubuntu-latest
-    steps:
-      - name: Notify Discord
+      - name: Send Discord notification
+        uses: 8398a7/action-slack@v2
+        if: failure()
```
neilzar: These jobs happen concurrently, so if both fail then you get double notifications.

.github/workflows/build.yml
```diff
@@ -0,0 +1,49 @@
+name: Build
+
+on:
+  push:
+  pull_request:
```
neilzar: It now runs twice each push if there is a PR open for the branch.

.github/workflows/build.yml
CreepPork: But they are different jobs, if Validation fails then Build may succeed.

To be frank, there are really two options:
1. This approach for notifications if one job fails (if you write really bad code that causes both jobs to fail then, yes, you'll get double notifications)
2. Continue to use Rezent/Mechilles which is an external application that has to be kept up to date too.

.github/workflows/build.yml
jonpas:
```suggestion
on:
  push:
    branches:
    - master
  pull_request:
```

CreepPork: There's still a small difference between the two currently "_duplicate_" jobs. Actions will build for the current branch and the other job will be built with the current branch merged into master.

neilzar: Even then, do you really need to do it twice? Jonpas' suggestion is perfect, no need to run it on every push. It only really becomes important whether it builds when it is merged into master, until then it is only secondary.

CreepPork:
For some reason, builds aren't triggering anymore.

mharis001 (approved):

PR #56, opened 2019-07-04
Add multi-line edit box control types

### LINKED ISSUE #48: Feature Request: Text Box
Ares had a `CopyPaste` dialog that was a huge text box used to copy or paste text from

mharis001:
- Add `EDIT:MULTI` and `EDIT:CODE` multi-line edit box sub-types to dialog
- Close #48

### DIFF
--- a/addons/dialog/functions/fnc_gui_edit.sqf
+++ b/addons/dialog/functions/fnc_gui_edit.sqf
@@ -19,10 +19,19 @@
 #include "script_component.hpp"
 
 params ["_controlsGroup", "_rowIndex", "_currentValue", "_rowSettings"];
-_rowSettings params ["_fnc_sanitizeValue"];
+_rowSettings params ["_fnc_sanitizeValue", "_isMulti", "_height"];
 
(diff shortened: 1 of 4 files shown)


addons/dialog/functions/fnc_gui_edit.sqf
```diff
@@ -19,11 +19,20 @@
 #include "script_component.hpp"
```
dedmen: You should consider putting the include before the documentation like in CBA/ACE. Otherwise it causes your line numbers to be messed up in case of errors.

mharis001: Yeah, need to make that change project wide.

CreepPork: To not add all the changes in this PR, I guess you could merge this one and then make a new PR to change it.

dedmen: yeah, but could atleast make sure that future PR's are already doing that.

PR #58, opened 2019-07-05
Add performance counters

mharis001:
- Add performance counters debug tool


addons/common/functions/fnc_dumpPerformanceCounters.sqf
```diff
@@ -0,0 +1,56 @@
...
+ * None
+ *
+ * Example:
+ * [] call zen_common_fnc_dumpPerformanceCounters
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+diag_log text "~~~ ZEN PERFORMANCE COUNTERS DUMP ~~~";
```
CreepPork: Couldn't you use the CBA macro `LOG`? And `TRACE_x` for all the others that have data?

addons/main/script_debug.hpp
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    #define COUNTER_NAME(x) GVAR(DOUBLES(x,counter))
+    #define CREATE_COUNTER(x) if (isNil "ZEN_COUNTERS") then {ZEN_COUNTERS = []}; COUNTER_NAME(x) = [QUOTE(COUNTER_NAME(x)), []]; ZEN_COUNTERS pushBack COUNTER_NAME(x)
+    #define BEGIN_COUNTER(x) if (isNil QUOTE(COUNTER_NAME(x))) then {CREATE_COUNTER(x)}; COUNTER_NAME(x) set [2, diag_tickTime]
+    #define END_COUNTER(x) (COUNTER_NAME(x) select 1) pushBack [COUNTER_NAME(x) select 2, diag_tickTime]
+#else
+    #define CREATE_COUNTER(x) /* disabled */
+    #define BEGIN_COUNTER(x) /* disabled */
+    #define END_COUNTER(x) /* disabled */
```
CreepPork: If disabled, then remove?

mharis001: What do you mean by remove? We still need the macros to be defined.

CreepPork: Whoops.

addons/common/functions/fnc_dumpPerformanceCounters.sqf
mharis001: That would add the formatting that those macros come with. Output is cleaner like this.

CreepPork: But you have `diag_logs` couple of lines down that use formats. It's cleaner to use the macros.

CreepPork: Plus, is there any reason that `text` is called? Doesn't look very structured text to me.

veteran29: diag_log "asdf" => `"asdf"` in file
diag_log text "asdf" => `asdf` in file

CreepPork: Ah, I see.

PR #66, opened 2019-07-07
Add full arsenal

oOKexOo:
- Add inventory module category
- Add `Add Full Arsenal` and `Remove Arsenal` modules


addons/modules/stringtable.xml
```diff
@@ -1170,5 +1170,23 @@
             <German>Es gibt keine Zielmodule, platzieren Sie zuerst ein Zielmodul</German>
             <Polish>Brak modułów celu, najpierw ustaw moduł celu</Polish>
         </Key>
+        <Key ID="STR_ZEN_Modules_ModuleAddFullArsenal">
```
mharis001: Move these above the feedback messages.

oOKexOo:
This goes down to the fundamental question how we want to handle integration of ACE features in general. The question will pop up again when adding Arsenal options to the inventory attribute window.

A) Provide Vanilla feature only.
B) Provide both simultaneously (Pretty much how Achilles does it, except when editing the loadout of units).
C) The choice is given by the CBA settings.

Also related to #69.

jonpas:
I would always go for ACE3 if ACE3 is running, otherwise vanilla.

As for modules specifically, if Zen ones are better or duplicate as ACE3, hide the ACE3 ones, if not hide the Zen ones. I'd even go as far to say we don't really need some modules in ACE3 that are currently there, but instead encourage people to use Zen, but we are where we are now.

CreepPork:
The settings are there for a reason and it states which arsenal is the preferred one. If the user has selected the one which he wants to use, then you shouldn't force something different on him. You wouldn't want to provide two identical modules but then their functionality is different. It doesn't make sense.

I'd say go what the user picked from the settings.

jonpas:
> You wouldn't want to provide two identical modules but then their functionality is different. It doesn't make sense.

So are you going to have a setting for each single module then? Otherwise you are n't solving the problem of different functionalities.

CreepPork:
> > You wouldn't want to provide two identical modules but then their functionality is different. It doesn't make sense.
> 
> So are you going to have a setting for each single module then? Otherwise you are n't solving the problem of different functionalities.

What are you on about? If there is a setting available that clearly states `Prefered Arsenal` then it should be used and shouldn't be limited just to one attribute button and then someone comes along and creates an issue saying that it ignores the setting. Your approach is forcing something upon the user that is absolutely not necessary.

jonpas:
I am talking about ACE integration in general, not only about Arsenal. But you prove my point, will you have settings for preferred module for others too? I am not opposed, but this can get out of hand quickly.

dedmen:
>will you have settings for preferred module for others too

If you mean whether to hide the ACE or the ZEN module for a certain thing, I think the developer can just decide that instead of letting the user configure it per module.

addons/modules/functions/fnc_moduleAddFullArsenal.sqf
```diff
@@ -0,0 +1,32 @@
+#include "script_component.hpp"
+/*
+ * Author: CreepPork_LV, Kex
```
CreepPork:
```suggestion
 * Author: CreepPork, Kex
```

addons/modules/functions/fnc_moduleRemoveArsenal.sqf
```diff
@@ -0,0 +1,35 @@
+#include "script_component.hpp"
+/*
+ * Author: CreepPork_LV, Kex
```
CreepPork:
```suggestion
 * Author: CreepPork, Kex
```
I tend to keep it to my GitHub username.

addons/modules/stringtable.xml
```diff
@@ -1070,6 +1070,24 @@
             <German>Spawn</German>
             <Polish>Spawn</Polish>
         </Key>
+        <Key ID="STR_ZEN_Modules_ModuleAddFullArsenal">
+            <Chinese>添加完整的軍火庫</Chinese>
+            <Chinesesimp>添加完整军火库</Chinesesimp>
+            <English>Add Full Arsenal</English>
```
mharis001: English at the top.

addons/modules/stringtable.xml
```diff
@@ -1070,6 +1070,24 @@
...
+            <French>Ajouter Arsenal complète</French>
+            <German>Komplettes Arsenal hinzufügen</German>
+            <Japanese>完全な武器庫を追加</Japanese>
+            <Russian>Полный Арсенал</Russian>
+        </Key>
+        <Key ID="STR_ZEN_Modules_ModuleRemoveArsenal">
+            <Chinese>移除軍火庫</Chinese>
+            <Chinesesimp>移除军火库</Chinesesimp>
+            <English>Remove Arsenal</English>
```
mharis001: English at the top.

addons/modules/functions/fnc_moduleRemoveArsenal.sqf
```diff
@@ -0,0 +1,35 @@
...
+    [LSTRING(NoObjectSelected)] call EFUNC(common,showMessage);
+};
+
+if (EGVAR(common,preferredArsenal) == 1 && {isClass (configFile >> "CfgPatches" >> "ace_arsenal")}) then {
+    [_object, true] call ace_arsenal_fnc_removeBox;
+} else {
+    // Set variables manually, since "AmmoboxExit" is not reliable
+    _object setVariable ["BIS_addVirtualWeaponCargo_cargo", nil, true];
+    private _boxes = missionNamespace getVariable ["BIS_fnc_arsenal_boxes", []];
+    _boxes = _boxes - [_object];
```
mharis001: Nitpick but don't need the local variable, can just do the `_boxes - [_object]` in the `setVariable` below.

addons/modules/config.cpp
```diff
@@ -4,6 +4,7 @@ class CfgPatches {
     class ADDON {
         name = COMPONENT_NAME;
         units[] = {
+            QGVAR(moduleAddFullArsenal),
```
mharis001: Maybe just `moduleAddArsenal` since there is only adding/removing full arsenals.

CreepPork: But down the road, a new module could be added for a limited arsenal and then you'd have to change this again.

addons/modules/functions/fnc_moduleRemoveArsenal.sqf
oOKexOo: Well, I could sure make an ugly one-liner just to get rid of the local variable `_boxes`, but why though?

addons/modules/config.cpp
oOKexOo: I guess you want to make a distinction between what kind of Arsenal is placed for the players and what kind of Arsenal is used when Zeus is editing a loadout.

addons/modules/functions/fnc_moduleRemoveArsenal.sqf
mharis001: Not about shortening code, more-so that I thought the code would look cleaner as:
```sqf
private _boxes = missionNamespace getVariable ["BIS_fnc_arsenal_boxes", []];
missionNamespace setVariable ["BIS_fnc_arsenal_boxes", _boxes - [_object], true];
```

addons/modules/config.cpp
oOKexOo: Sry, my last comment did not fit here.
I would keep the name `moduleAddFullArsenal` based on @CreepPork argument.

PR #70, opened 2019-07-08
Add sensors button to attributes displays

mharis001:
- Add sensors button to vehicle attributes displays to control vehicle data link capabilities


mharis001:
It is in the `garage` component, I moved it there but forgot to remove the entry from `attributes`.

PR #75, opened 2019-07-14
Show custom modules in recent tree

### LINKED ISSUE #74: Feature Request: Custom modules show up in the "Recent" tab
Not sure if it is possible, tho i would be pretty cool if custom modules would show up in the "Recent" (star) tab after use like all the other modules.

mharis001:
- Show used custom modules in the recent tree
- Looks like recent tree ignores `scopeCurator = 1` modules
- Close #74


addons/custom_modules/CfgVehicles.hpp
```diff
@@ -4,8 +4,9 @@ class CfgVehicles {
     #define ADD_CUSTOM_MODULE(id) \
         class DOUBLES(GVAR(module),id): EGVAR(modules,moduleBase) { \
             displayName = CSTRING(DisplayName); \
+            category = QGVAR(category); \
```
CreepPork: Is it needed here? And was there an issue before this was added?

addons/custom_modules/CfgVehicles.hpp
```diff
@@ -4,8 +4,9 @@ class CfgVehicles {
...
             function = QFUNC(init); \
-            scopeCurator = 1; \
+            scopeCurator = 2; \
             GVAR(index) = __EVAL(id - 1); \
```
CreepPork: Why bother with removing 1? Does it cause any behaviour if the id is zero-based? To my eyes, I can't spot anything, apart from the classname (which wouldn't be any issue if it would be zero-based?)

addons/custom_modules/CfgVehicles.hpp
mharis001: Yes, needed. No.

addons/custom_modules/CfgVehicles.hpp
mharis001: No either would work, just nicer having the class names go from 1-100.

PR #79, opened 2019-07-15
Add change weather module

### LINKED ISSUE #30: Feature Request: Advanced Weather Change
Instead of the Vanilla Zeus module in which you can just apply rain, overcast suddenly. There should be a module to allow for a smoother transition over to new weather. Such as increase in rain, overcast increases over a period of time. 
Advanced Weather Change

mharis001:
- Add change weather module, Close #30 
- Change dialog `SLIDER` control's decimal argument to formatting
    - Allows custom formatting code (example: degree symbol)

### DIFF
--- a/addons/modules/stringtable.xml
+++ b/addons/modules/stringtable.xml
@@ -1187,6 +1187,75 @@
             <German>Anzahl der Schüsse, die von jeder Einheit abgegeben werden sollen.</German>
             <Polish>Liczba pocisków do wystrzelenia przez każdą jednostkę.</Polish>
         </Key>
+        <Key ID="STR_ZEN_Modules_ModuleWeather">
+            <English>Change Weather</English>
+        </Key>
+        <Key ID="STR_ZEN_Modules_ModuleWeather_Overcast">
+            <English>Overcast</English>
+        </Key>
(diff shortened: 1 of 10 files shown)


addons/modules/stringtable.xml
```diff
@@ -1187,6 +1187,75 @@
...
+            <English>Sets the wind speed in kilometers per hour.</English>
+        </Key>
+        <Key ID="STR_ZEN_Modules_ModuleWeather_WindDirection">
+            <English>Wind Direction</English>
+        </Key>
+        <Key ID="STR_ZEN_Modules_ModuleWeather_WindDirection_Tooltip">
+            <English>Sets the wind direction.</English>
+        </Key>
+        <Key ID="STR_ZEN_Modules_ModuleWeather_Gusts">
+            <English>Gusts</English>
```
CreepPork: Wind Gusts? I haven't seen anyone using just gusts.

mharis001:
No, but I think ACE Weather only does `setWind` now and either way don't think there is any good way to provide compatibility for that. It is one or the other, depends on how the mission is setup - also requires the manual override to be checked.

addons/modules/stringtable.xml
mharis001: 3DEN has just "Gusts Start" and "Gusts Forecast".

PR #80, opened 2019-07-16
Improve dialog toolbox control

mharis001:
- Improve dialog toolbox control with support any number of rows and columns
- Add a wide variant - `TOOLBOX:WIDE`


mharis001:
You could but, I think there are some problems that it not worth doing. For example, you would need a max column count or how do you handle longer/shorter strings with that. It is simpler and easier for the user to specify exactly what type of toolbox they want.

In terms of BWC, I guess a check can be added for the old value info format and then convert over.

PR #83, opened 2019-07-20
Add waypoint types config

mharis001:
- Add config to easily configure waypoint types available to Zeus
- Attribute UI will scale automatically as more are added

### DIFF
--- a/addons/attributes/CfgWaypointTypes.hpp
+++ b/addons/attributes/CfgWaypointTypes.hpp
@@ -1,0 +1,52 @@
+class ZEN_WaypointTypes {
+    class Move {
+        displayName = "$STR_ac_move";
+        type = "MOVE";
+    };
+    class Cycle {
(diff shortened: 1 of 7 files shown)


bux:
Kinda missing "GUARD" waypoint, isn't it?

mharis001:
It's easy to add the waypoint type but since you cannot create/interact with triggers in Zeus, you would not be able to make use of it. I am going to look into using the `createGuardedPoint` command with a module, so Zeus' can create custom guard points through that.

mharis001:
After some investigation, I don't think guard waypoints are going to be possible through Zeus currently. `createGuardedPoint` creates a guard point that cannot be move and is unable to be deleted and creating a trigger of the "Guarded by" type does not seem to work.

Would need the `createGuardedPoint` command to be updated to return something that represents the guard point so that it can be deleted or moved. Or best would be if an object is supplied the point moves with the object.

PR #86, opened 2019-07-25
Expand player/zeus teleport context actions with teleport to vehicle capability

3Mydlo3:
- Expand player/zeus teleport context actions with teleport to vehicle capability

If action is used when hovered over a vehicle it will try to fit all selected players (or just Zeus) into the vehicle (taking into account that some of them may already be inside). If it fails to find enough space, message is shown and action does nothing to prevent accidental teleportation.


addons/common/XEH_PREP.hpp
```diff
@@ -19,3 +19,4 @@ PREP(setTurretAmmo);
 PREP(setVehicleAmmo);
 PREP(showMessage);
 PREP(spawnLargeObject);
+PREP(teleportToVehicle);
```
bux: `teleportIntoVehicle`?

addons/context_actions/functions/fnc_teleportPlayers.sqf
```diff
@@ -21,8 +21,17 @@ private _players = _this select {isPlayer _x};
...
-        {
-            _x setPosASL _posASL;
-        } forEach _players;
+        private _curatorHovered = curatorMouseOver;
+        _curatorHovered params ["_type", "_entity"];
+
+        if (_type isEqualTo "OBJECT") then {
+            private _result = [_players, _entity] call EFUNC(common,teleportToVehicle);
+            if (_result) exitWith {};
+            [localize "str_3den_notifications_vehiclefull_text"] call EFUNC(common,showMessage);
```
mharis001: The message is: "No space in target vehicle" but I think this will show when there are not enough empty positions to fill all units (trying to teleport 2 people into a vehicle with 1 free seat). Maybe we should use a custom string?

Also, don't need `localize`, `showMessage` handles that.

3Mydlo3: If we want custom string we might use something like "Not enough space in target vehicle: %1 needed - %2 available" so Zeus will get an information about target vehicle cargo space which might be useful. Zeus will know exactly how much space there is instead making blind guesses or putting players one by one.

addons/common/XEH_PREP.hpp
3Mydlo3: Yeah, that sounds much better.

addons/common/functions/fnc_teleportToVehicle.sqf
```diff
@@ -0,0 +1,35 @@
...
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_units", "_vehicle"];
+
+private _success = false;
+
+if (_vehicle isKindOf "AllVehicles" && {!(_vehicle isKindOf "CAManBase")}) then {
+    private _freeCargoSpace = {_x#0 isEqualTo objNull} count fullCrew [_vehicle, "", true];
```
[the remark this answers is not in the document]
3Mydlo3: Is it project-wide restriction? I know that you follow ACE Coding Guidelines but not everything can be found there. Maybe there is something more that I'm not aware of.

addons/context_actions/functions/fnc_teleportPlayers.sqf
mharis001: 👍 Sounds good.

addons/common/functions/fnc_teleportToVehicle.sqf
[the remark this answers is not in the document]
mharis001: Yeah, the rest of the project does not use `#`, always `select`. I don't think the ACE guidelines say anything about this, just something for consistency with the rest of the code.

addons/context_actions/functions/fnc_teleportZeus.sqf
```diff
@@ -17,4 +17,10 @@
 
 params ["_posASL"];
 
-player setPosASL _posASL;
+curatorMouseOver params ["_type", "_entity"];
+
+if (_type isEqualTo "OBJECT") then {
```
3Mydlo3: Right now when teleport action is used when hovered over any object nothing will happen if it is not a vehicle. I think it would be better to check here for a vehicle instead of checking that inside `teleportIntoVehicle`.

mharis001: Agreed. 👍

addons/common/functions/fnc_teleportIntoVehicle.sqf
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Return Value:
+ * Success <BOOL>
+ *
+ * Example:
+ * [[player1, player2, player3], vehicle player] call zen_common_fnc_teleportToVehicle
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
```
mharis001: Move to the top of the file, reference: #84 .

addons/common/functions/fnc_teleportIntoVehicle.sqf
```diff
@@ -0,0 +1,38 @@
...
+    private _unitsNotInTargetVehicle = _units select {!(_x in _vehicle)};
+    private _unitsCount = count _unitsNotInTargetVehicle;
+
+    if (_unitsCount <= _freeCargoSpace) then {
+        {
+            _x moveInAny _vehicle;
+        } forEach _unitsNotInTargetVehicle;
+        _success = true;
+    } else {
+        [ELSTRING(common,VehicleNotEnoughSpace_Xneeded_Yavailable), _unitsCount, _freeCargoSpace] call EFUNC(common,showMessage);
```
mharis001: This is in the common component, can use just `LSTRING`. Maybe shorten the stringtable key name, something like `NotEnoughSeatsInVehicle` or `NotEnoughSpaceInVehicle`.

addons/context_actions/functions/fnc_teleportZeus.sqf
```diff
@@ -13,8 +14,13 @@
...
  */
-#include "script_component.hpp"
 
 params ["_posASL"];
 
-player setPosASL _posASL;
+curatorMouseOver params ["_type", "_entity"];
+
+if (_type isEqualTo "OBJECT" && {_entity isKindOf "AllVehicles"} && {!(_entity isKindOf "CAManBase")}) then {
+    private _result = [[player], _entity] call EFUNC(common,teleportIntoVehicle);
```
mharis001: Remove `_result` variable.

3Mydlo3: I don't know how I missed that.

PR #91, opened 2019-08-01
Use hovered entity for teleport Zeus

mharis001:
- Use `_hoveredEntity` instead of `curatorMouseOver` for Teleport Zeus action
- `curatorMouseOver` can change while `_hoveredEntity` will remain as the entity from when the menu was opened

### DIFF
--- a/addons/context_actions/functions/fnc_teleportZeus.sqf
+++ b/addons/context_actions/functions/fnc_teleportZeus.sqf
@@ -4,13 +4,13 @@
  * Teleports Zeus to the given position or vehicle.
  *
  * Arguments:
- * 0: Position <ARRAY>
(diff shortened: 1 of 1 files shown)


bux:
Just curious: where is the var `_hoveredEntity` assigned?

mharis001:
Here: 

https://github.com/zen-mod/ZEN/blob/026981981087d5c8220b919eeaada98d44542b40/addons/context_menu/functions/fnc_createContextGroup.sqf#L107-L110

and the macro is:

https://github.com/zen-mod/ZEN/blob/026981981087d5c8220b919eeaada98d44542b40/addons/context_menu/script_component.hpp#L40-L43

PR #93, opened 2019-08-02
Add OWNERS control and improve Teleport Players module

mharis001:
- Add `OWNERS` control type to dialog
    - Returns the selected tab, sides, groups, players
    - Can select multiple sides, groups, players
- Add `getPlayers` common function
    - Returns list of players based on given sides, groups, and units
- Improve Teleport Players module
    - Better teleporting behaviour
    - Add Include vehicles option
    - Remembers previous selection now


addons/common/functions/fnc_getPlayers.sqf
```diff
@@ -0,0 +1,24 @@
...
+ * Public: No
+ */
+
+params [["_entities", [], [objNull, grpNull, west, []]]];
+
+if !(_entities isEqualType []) then {
+    _entities = [_entities];
+};
+
+call CBA_fnc_players select {_x in _entities || {group _x in _entities} || {side group _x in _entities}}
```
CreepPork: Doesn't the project use `[] call ...`? I recall seeing that more.

addons/dialog/functions/fnc_gui_owners.sqf
```diff
@@ -0,0 +1,234 @@
...
+    } forEach IDCS_ROW_OWNERS_BTNS;
+
+    {
+        private _ctrlTab = _controlsGroup controlsGroupCtrl _x;
+        _ctrlTab ctrlShow (_forEachIndex == _selectedTab);
+    } forEach IDCS_ROW_OWNERS_TABS;
+
+    ctrlSetFocus (_controlsGroup controlsGroupCtrl (IDC_ROW_OWNERS_TABS select _selectedTab));
+
+    _currentValue set [0, _selectedTab + 1]; // +1 since one-based index
```
CreepPork: Why `+1 since one-based index`? It seems counter-intuitive of a comment. If it was one-based then you would subtract one not add one.

mharis001: More so that the index used by the control type itself (1 - Sides, 2 - Groups, 3 - Players). I wasn't sure on this but it does enable:

```sqf
_values params ["_owners"];
_owners = _owners select (_owners select 0);
```

Could change to 0-2 or move to last argument in value info.

mharis001:
Should be fixed now, using `moveOut` and then a frame later `setVehiclePosition`.

PR #106, opened 2019-08-06
Fix artillery fire mission spread behaviour

### LINKED ISSUE #99: Artillery spread works more like an offset
https://gyazo.com/19cfe5d3890d36fc9b7597c371e4292b
https://gyazo.com/8f0a3492e7300ab241d51b871995d2e4

mharis001:
- title, Fix #99 
- Fire each round individually at a random position around the center
- Rounds are not fired as fast but produces the correct behaviour that "spread" implies

### DIFF
--- a/addons/modules/XEH_postInit.sqf
+++ b/addons/modules/XEH_postInit.sqf
@@ -50,6 +50,12 @@ if (isServer) then {
 [QGVAR(fireArtillery), {
     params ["_unit", "_position", "_spread", "_ammo", "_rounds"];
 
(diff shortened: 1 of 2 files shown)


mharis001:
Yeah, that is the result of having to use `doArtilleryFire` with one round at a time, I don't think there is another way. 

Could just use a single `doArtilleryFire` when spread is low (<10) since that should produce the expected result in that situation as well. Will help avoid this when using a small spread value, or none.

oOKexOo:
Yeah, when no spread is selected, just use `doArtilleryFire` once.

PR #108, opened 2019-08-06
Move camera to unit's position when exiting remote control

### LINKED ISSUE #103: Option to set camera position on the unit's position when remote control is released
It could be a very useful feature as you might want to be in the same area as the unit you just controlled.

mharis001:
- title, Close #103

### DIFF
--- a/addons/remote_control/functions/fnc_start.sqf
+++ b/addons/remote_control/functions/fnc_start.sqf
@@ -22,10 +22,11 @@ _unit = effectiveCommander _unit;
 _unit setVariable [VAR_OWNER, player, true];
 missionNamespace setVariable [VAR_UNIT, _unit];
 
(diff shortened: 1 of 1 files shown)


mharis001:
What do you mean by "same level of zoom"?

CreepPork:
I'll post a GIF later, but the camera doesn't keep the same distance to the unit (your field of view). It points the camera at the ground where only the ground and the unit is visible. It's better than nothing but it would be nice if it would at least increase its distance. There would be a problem taking control of planes and then releasing it though.

mharis001:
The current 10m back, 10m up from the unit is the same as BI's remote control. What would a better distance be?

CreepPork:
Probably the camera angle would be a great way to start as it’s aligned too much to the ground. And for the distance, maybe a toggleable setting, that keeps the same distance from the unit when the RC was started, only the position of the camera is moved.

PR #114, opened 2019-08-08
Add blank component

mharis001:
- Add blank template for easier creation of new components

### DIFF
--- a/extras/blank/config.cpp
+++ b/extras/blank/config.cpp
@@ -1,0 +1,17 @@
+#include "script_component.hpp"
+
+class CfgPatches {
+    class ADDON {
+        name = COMPONENT_NAME;
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"zen_common"};
+        author = ECSTRING(main,Author);
+        authors[] = {};
+        url = ECSTRING(main,URL);
+        VERSION_CONFIG;
+    };
+};
+
+#include "CfgEventHandlers.hpp"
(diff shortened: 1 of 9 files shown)


extras/blank/stringtable.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project name="ZEN">
+    <Package name="Blank">
+
```
CreepPork: Maybe add a demo line?

extras/blank/config.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+class CfgPatches {
+    class ADDON {
+        name = COMPONENT_NAME;
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"zen_common"};
+        author = ECSTRING(main,Author);
+        authors[] = {};
```
CreepPork: Maybe add a demo author? Otherwise, you could skip that while quickly going through.

mharis001: Not what that would be and I think leaving it empty is better than something like `authors[] = {"blank_author"};`.

CreepPork:
```suggestion
        authors[] = {"Username"};
```

extras/blank/config.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+class CfgPatches {
+    class ADDON {
+        name = COMPONENT_NAME;
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"zen_common"};
+        author = ECSTRING(main,Author);
+        authors[] = {"username"};
```
CreepPork: Capitalized:
```suggestion
        authors[] = {"Username"};
```

PR #115, opened 2019-08-08
Add missing music tracks from the vanilla game into the Play Music module

CreepPork:
- Add missing tracks
- Reorganize some of the existing ones so it stays more organized and makes more sense


addons/music/script_component.hpp
```diff
@@ -0,0 +1,17 @@
+#define COMPONENT music
+#define COMPONENT_BEAUTIFIED Music
+#include "\x\zen\addons\main\script_mod.hpp"
+
+// #define DEBUG_MODE_FULL
+// #define DISABLE_COMPILE_CACHE
+// #define ENABLE_PERFORMANCE_COUNTERS
+
+#ifdef DEBUG_ENABLED_MARKERS_TREE
```
mharis001: Fix macro for this component.

addons/music/CfgMusic.hpp
```diff
@@ -0,0 +1,124 @@
+class CfgMusic {
+    // This Is War a
+    class LeadTrack01a_F {
+        musicClass = "Lead";
+        name = "$STR_ZEN_Music_ThisIsWarPartOne";
```
mharis001: Use `CSTRING` here and below.

addons/music/XEH_PREP.hpp
```diff
@@ -0,0 +1 @@
+//
```
mharis001: Remove?

PR #119, opened 2019-08-10
Add paradrop waypoint

mharis001:
- Add scripted paradrop waypoint

### DIFF
--- a/addons/ai/functions/fnc_waypointParadrop.sqf
+++ b/addons/ai/functions/fnc_waypointParadrop.sqf
@@ -17,6 +17,8 @@
  */
 
 #define MOVE_DELAY 3
+#define OFFSET_DISTANCE 2000
+#define TIME_PER_UNIT 1.2
 
 params ["_group", "_waypointPosition"];
 
@@ -35,6 +37,7 @@ private _skill = skill _driver;
 _driver setSkill 1;
 _driver allowFleeing 0;
 
+private "_position";
 private _nextMove = CBA_missionTime;
 
 waitUntil {
@@ -42,7 +45,7 @@ waitUntil {
     // Makes the aircraft fly over the waypoint's position and not stop at it
     if (CBA_missionTime >= _nextMove) then {
         private _direction = _vehicle getDir _waypointPosition;
-        private _position = _waypointPosition getPos [2000, _direction];
+        _position = _waypointPosition getPos [OFFSET_DISTANCE, _direction];
         _vehicle move _position;
 
         _nextMove = CBA_missionTime + MOVE_DELAY;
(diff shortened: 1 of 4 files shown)


addons/ai/functions/fnc_waypointParadrop.sqf
```diff
@@ -0,0 +1,63 @@
...
+private _driver = driver _vehicle;
+private _skill = skill _driver;
+
+// Increase the skill of the pilot for better flying
+_driver setSkill 1;
+_driver allowFleeing 0;
+
+private _nextMove = CBA_missionTime;
+
+waitUntil {
```
CreepPork: Per ACE3 coding guidelines it is disallowed to use it.

mharis001: Scripted waypoints have to be scheduled.

addons/ai/functions/fnc_waypointParadrop.sqf
```diff
@@ -0,0 +1,63 @@
...
+
+        _nextMove = CBA_missionTime + MOVE_DELAY;
+    };
+
+    sleep 0.5;
+
+    // Check if the aircraft is close enough to the waypoint to begin paradropping
+    // Distance is based on the current speed of aircraft and the number of passengers
+    private _passengers = {assignedVehicleRole _x select 0 == "cargo"} count crew _vehicle;
+    _vehicle distance2D _waypointPosition < vectorMagnitude velocity _vehicle * 1.5 * _passengers / 2 max 50
```
oOKexOo: 1) I think it should be `1.2` instead of `1.5`, since `ZEN_fnc_ejectPassengers` has 1.2 s as delay between the passengers.
If you want to account for the 1.5 s before opening the chute, you could add  `_vehicleSpeed * 1.5` in addition.
Anyway, we can improve the trajectories later on by doing displacement measurements like I did for HALO once.
2) Also consider measuring the distance between `_position` and `_vehicle` rather than `_waypointPosition` and `_vehicle`. That way you don't need the set the minimum to 50 m.

mharis001: 1. Can try 1.2 instead of 1.5.
2. `_position` is 2 km past `_waypointPosition` in the direction from the aircraft to `_waypointPosition`. This would cause the units to eject late and not over the waypoint's position.

oOKexOo: I mean something like
```suggestion
_vehicle distance2D _position <  (vectorMagnitude velocity _vehicle) * (1.2 * _passengers / 2 + 1.5) + 2000;
```
The offsets we use are in respect to the direction of travel, which is better described by the vector `_position vectorDiff getPos _vehicle` than  `_waypointPosition vectorDiff getPos _vehicle`. Hence, you will get better results in the case the aircraft does not exactly pass the target location.

CreepPork: Also, the numbers should be put into preprocessor defines as per the no magic numbers rule.

PR #120, opened 2019-08-11
Add Area Markers

mharis001:
- Add ability to create area markers with Zeus
- Can be created by either using the module or context menu on the map
- Can be moved around by clicking on the center icon and dragging
- Properties can be edited by double clicking on the center icon
- Can be deleted by hovering over the center icon and pressing delete key
- One small annoyance is pressing escape to close the configure menu also closes the map
    - No workaround, hard coded engine behaviour (ignores handled KeyDown EH)
    - Can't manually use `ctrlShow` on the map control either
- Improve `initSliderEdit` common function with support for percentage sliders and CODE as formatting parameter

<details>
<summary>Image</summary>

area_markers

</details>

### DIFF
--- a/addons/area_markers/XEH_postInit.sqf
+++ b/addons/area_markers/XEH_postInit.sqf
@@ -1,0 +1,84 @@
+#include "script_component.hpp"
+
+if (isServer) then {
+    [QGVAR(create), {
+        params ["_position"];
+
+        private _marker = createMarker [format [QGVAR(%1), GVAR(nextID)], _position];
+        _marker setMarkerShape "RECTANGLE";
+        _marker setMarkerSize [50, 50];
+
+        GVAR(markers) pushBack _marker;
+        publicVariable QGVAR(markers);
+
+        GVAR(nextID) = GVAR(nextID) + 1;
+
+        [QGVAR(createIcon), _marker] call CBA_fnc_globalEvent;
+    }] call CBA_fnc_addEventHandler;
+
+    [QGVAR(delete), {
+        params ["_marker"];
+
+        private _index = GVAR(markers) find _marker;
+
+        if (_index != -1) then {
+            GVAR(markers) deleteAt _index;
+            publicVariable QGVAR(markers);
+
+            deleteMarker _marker;
+
+            [QGVAR(deleteIcon), _marker] call CBA_fnc_globalEvent;
+        };
+    }] call CBA_fnc_addEventHandler;
+};
+
+if (hasInterface) then {
+    ["ZEN_displayCuratorLoad", {
+        params ["_display"];
+
+        // Namespace of marker names and their corresponding icon controls
+        if (isNil QGVAR(icons)) then {
+            GVAR(icons) = [] call CBA_fnc_createNamespace;
+        };
+
+        // Add EH to update area marker icon positions when the map is shown
+        private _ctrlMap = _display displayCtrl IDC_RSCDISPLAYCURATOR_MAINMAP;
+        _ctrlMap ctrlAddEventHandler ["Draw", {call FUNC(update)}];
+
+        // Add EH to handle deleting area marker by pressing the DELETE key
+        _display displayAddEventHandler ["KeyDown", {call FUNC(onKeyDown)}];
+
+        // Create area marker icons for all area markers
+        {
+            {
+                [_x] call FUNC(createIcon);
+            } forEach GVAR(markers);
+        } call CBA_fnc_execNextFrame;
+
+        // Add PFH to update visibility of area marker icons
+        GVAR(visiblePFH) = [{
+            params ["_args"];
+            _args params ["_visible"];
+
+            if (_visible isEqualTo visibleMap) exitWith {};
+
+            _visible = visibleMap;
+            {
+                private _ctrlIcon = GVAR(icons) getVariable [_x, controlNull];
+                _ctrlIcon ctrlShow _visible;
+            } forEach GVAR(markers);
+
+            _args set [0, _visible];
+        }, 0, [visibleMap]] call CBA_fnc_addPerFrameHandler;
+    }] call CBA_fnc_addEventHandler;
+
+    ["ZEN_displayCuratorUnload", {
+        GVAR(visiblePFH) call CBA_fnc_removePerFrameHandler;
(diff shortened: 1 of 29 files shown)


addons/area_markers/XEH_postInit.sqf
```diff
@@ -0,0 +1,84 @@
+#include "script_component.hpp"
+
+if (isServer) then {
```
CreepPork: Perhaps this could be extracted into a separate file and add a doc block as this is quite long?

addons/area_markers/functions/fnc_configure.sqf
```diff
@@ -0,0 +1,193 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * ["marker_0"] call zen_area_markers_fnc_configure
+ *
+ * Public: No
+ */
+
+#define DEGREE_SYMBOL (toString [176])
```
CreepPork: Isn't this defined for the second time? I think it was defined in the dynamic dialog slider as well. Perhaps extract it to common?

addons/area_markers/functions/fnc_module.sqf
```diff
@@ -0,0 +1,29 @@
...
+ * Zeus module function to create an area marker.
+ *
+ * Arguments:
+ * 0: Logic <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [LOGIC] call zen_area_markers_fnc_module
```
CreepPork: Seems a bit weird to even have a module. I think that the context menu would suffice or if possible double click on the map could be even better.

mharis001: Not everyone uses the context menu, it can be disabled.

CreepPork: But a double click on the map seems a better alternative than a context-based module that takes up space in the tree.

addons/area_markers/XEH_postInit.sqf
mharis001: I don't see what that would do, Just split up code when `XEH_postInit` is generally the first place to look for added CBA events. Also, I think the code is simple enough and doesn't need a doc block.

addons/area_markers/functions/fnc_module.sqf
CreepPork: Out of Zeus, if you double click on the map it creates a new marker and in Zeus you could use the same behaviour (only you create an area marker, perhaps could be extended later to allow all kinds of markers?). As for the attributes that would be only valid if we would be double-clicking on an object but in this instance that isn't the case.

mharis001: Attributes are for all Zeus entities (objects, groups, waypoints, markers). There is no method that I know of to create a Zeus editable marker.

I just don't see the need to make this accessible via double click when it's available through a right-click + left-click (context menu). Especially when it is such a niche feature.

CreepPork:
Is it possible to rotate the marker after it has been created?

addons/area_markers/functions/fnc_module.sqf
CreepPork: The issue is that the module is profoundly context-sensitive (e.g. you need the map to be open), whereas all the other modules are more about asking `place me on an object`.
And as the project evolves, the more of these modules you have, the bigger mess you are going to have of the module tree, especially if they are such a niche item as you said. What are you going to keep doing down the line? Add more of these types of modules? It doesn't scale.

Achilles has a module filter, but, I personally, never used it because it was too much of a hassle to use, especially if you just needed that module that was locked away. So that too isn't a viable solution to remove the module if you don't want it.

You already provide the option to edit and delete via the map without using the context menu then it would seem completely intuitive that you could create the marker via double click as well.

I'm not talking about adding attributes to the area markers, you're talking about adding something to the mix that isn't required right now.

I'm not saying that this should be restricted to only area markers. It could be later extended to allow usage of any markers not just area ones. Then you wouldn't have to deal with using that marker tree. This would drastically boost its potential later.

mharis001:
Yes, you can double click it to configure its properties (size, rotation, brush, etc).

addons/area_markers/functions/fnc_module.sqf
mharis001: > Then you wouldn't have to deal with using that marker tree.

The markers tree is the easiest way to create a marker (especially after the integration of the `markers_tree` component), I don't see how any other solution would be more straight forward. And in either case, there is no way to create a Zeus editable marker through script.

Any properties for the regular (icon) Zeus markers would be go through attributes, it does not make sense to have all other Zeus entity types be configured through attributes but markers through something separate. Right now, this UI is specific to area markers (shape, brush), trying to accommodate into one or selectively display elements does not sound like a good idea.

> I'm not talking about adding attributes to the area markers, you're talking about adding something to the mix that isn't required right now.

That is not what I said. I said that attributes are for "regular" Zeus entities and area markers are not one them.

> The issue is that the module is profoundly context-sensitive (e.g. you need the map to be open), whereas all the other modules are more about asking place me on an object.

The module needs to be placed on the map because that is what makes. It could easily be made to just take the position of the module when not on the map. However, that doesn't make sense - markers are a map related entity and it makes sense for interactions with them to go through the map.

> And as the project evolves, the more of these modules you have, the bigger mess you are going to have of the module tree, especially if they are such a niche item as you said. What are you going to keep doing down the line? Add more of these types of modules? It doesn't scale.

I don't see the module tree becoming a mess as long as each module provides a unique feature and is logically categorized. The number of items in the tree will grow as features are added, the only way you avoid that is through the exclusion of said features.

> You already provide the option to edit and delete via the map without using the context menu then it would seem completely intuitive that you could create the marker via double click as well.

I get what you are saying - on the regular "infantry" map, you create markers by double-clicking and I can see an argument for mimicking that on the Zeus map. However, this is only for area markers, as I said earlier, we can't create the icon Zeus markers through script. And in that case, for just area markers, the context menu is a simple out-of-the-way solution to provide this functionality (and the module if someone has the context menu disabled). We also currently have double-click to edit, and CTRL + double-click to remote control.

This is something that can always be changed/integrated later based on feedback. Personally, I don't see a need for it.

PR #123, opened 2019-08-14
Add Amount controls for flyby module, fix #121

### LINKED ISSUE #121: Feature Request: Add option to spawn multiple vehicles for Ambient Flyby
e.g. when wanting to have a formation of two F-18 for the flyby.

bux:
- Will add a control section for the ambient flyby module
- A user can change the amount from 1 to up to 6 aircraft (it's enough for an upper limit, isn't it?)


mharis001:
Since these are discrete values, maybe a toolbox control would be better than a slider?

CreepPork:
The toolbox would add too much stuff, a simple slider should keep things simple and elegant.

bux:
I'll continue with the slider control and if it isn't desired after all, it shouldn't be too hard to switch to a different one.

mharis001:
A slider and edit box is not simpler, I tried both out and a toolbox is visually superior, unless you wanted more than 10 but I think the current 6 is a reasonable maximum.

I don't know what the intended formation is but from me trying it out, they fly in a column (more-or-less). For that, maybe this should spawn them behind each other with respect to the flight direction.

bux:
> For that, maybe this should spawn them behind each other with respect to the flight direction.

I tried that and the diagonal shift seems to work better especially with the forward velocity on spawn (less crashes). Default formation is "Wedge".

CreepPork:
In aviation you’d never fly in a column due to jet wash and it overall being a more complicated procedure. Formation should be the wedge.

bux:
Would it make sense to add the spawned planes to each curator?

CreepPork:
Probably, you’d be able to keeep track of the aircraft better. Same behaviout in the CAS modules.

bux:
Btw. what is a toolbox control? Are we talking about a dropdown?

mharis001:
Yes, same as the directions control - `ctrlToolbox`.

bux:
This is pretty much done

PR #129, opened 2019-08-20
Add Deploy Countermeasures keybind

### LINKED ISSUE #116: Feature Request: Throw countermeasures from vehicles keybind (units throw smoke)
This would essentially be the same behaviour as found in Achilles. You press a keybind on a selected vehicle/unit and it deploys countermeasures/throws a smoke grenade.

mharis001:
- Add "Deploy Countermeasures" keybind, Close #116 
    - Makes all currently selected vehicles deploy smokes/flares
- Prevent activating other keybinds while searching

### DIFF
--- a/addons/common/functions/fnc_deployCountermeasures.sqf
+++ b/addons/common/functions/fnc_deployCountermeasures.sqf
@@ -17,7 +17,7 @@
 
 params ["_vehicle"];
 
-if !(_vehicle isKindOf "AllVehicles") exitWith {};
+if !(_vehicle isKindOf "LandVehicle" || {_vehicle isKindOf "Air"} || {_vehicle isKindOf "Ship"}) exitWith {};
 
 private _weapons = [];
 private _cfgAmmo = configFile >> "CfgAmmo";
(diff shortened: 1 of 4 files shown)


addons/common/functions/fnc_deployCountermeasures.sqf
```diff
@@ -0,0 +1,46 @@
...
+                    _weapons pushBackUnique _x;
+                };
+            } forEach (_vehicle weaponsTurret _turretPath);
+        };
+    };
+} forEach magazinesAllTurrets _vehicle;
+
+if (_weapons isEqualTo []) exitWith {};
+
+[_vehicle, selectRandom _weapons] call BIS_fnc_fire;
```
CreepPork: It returns a bool value in SP and nothing in MP.

CreepPork:
What about vehicles that have multiple types of countermeasures? E.g. RHS USAF's littlebird with the ECM?

CreepPork:
Would be very nice if this would be also applied to units.

mharis001:
I think this better kept to just vehicles. Think the grenades actions will be better for units or a separate keybind for units to throw smokes be added later.

addons/common/functions/fnc_deployCountermeasures.sqf
mharis001: Not really needed, but I added a `nil` at the end.

CreepPork: It's that just the doc block didn't correctly reflect the actual return value.

PR #130, opened 2019-08-22
Cleanup tabs in music component

mharis001:
- Cleanup tabs in music component `CfgMusic.hpp`


jonpas:
VSCode's SQF extension is breaking EditorConfig when opened with keybind I believe, @bux can elaborate.

PR #135, opened 2019-08-26
Visibility indicator tweaks

mharis001:
- Move setting out of subcategory and change default to false
    - Was the only setting in the subcategory
    - Seems more users will have it turned off
- Disable drawing while the interface is in screenshot mode
    - Add `isInScreenshotMode` common function
- Move the "VISIBLE" text closer to the cursor (less vertical space)
- Don't draw an indicator for the local player (thoughts?)

### DIFF
--- a/addons/visibility/functions/fnc_draw.sqf
+++ b/addons/visibility/functions/fnc_draw.sqf
@@ -19,10 +19,10 @@
 private _pos = AGLtoASL screenToWorld getMousePosition;
 private _intersections = lineIntersectsSurfaces [getPosASL curatorCamera, _pos];
 if !(_intersections isEqualTo []) then {
-    _pos = ((_intersections select 0) select 0);
+    _pos = _intersections select 0 select 0;
 };
 
-// check 1.5 above the cursor to prevent a small object on the terrain blocking the view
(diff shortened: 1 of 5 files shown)


addons/visibility/functions/fnc_draw.sqf
```diff
@@ -15,24 +15,28 @@
...
+    _pos = _intersections select 0 select 0;
 };
 
-// check 1.5 above the cursor to prevent a small object on the terrain blocking the view
+// Check 1.5 above the cursor to prevent a small object on the terrain blocking the view
 private _posHigh = _pos vectorAdd [0, 0, 1.5];
 private _draw = false;
 {
-    if (side _x != sideLogic && {(((_x getRelDir _posHigh) + 90) mod 360) < 180}) then {
+    if (_x != player && {side _x != sideLogic} && {((_x getRelDir _posHigh) + 90) mod 360 < 180}) then {
```
CreepPork: Seems like a lot of stuff is going in this singular if statement, maybe a comment is in order?

neilzar: @mharis001 Do you think a comment is necessary? I don't think it is,

mharis001: Doesn't hurt to add one.

PR #144, opened 2019-08-28
Add Custom Compositions

### LINKED ISSUE #295: Deep copy units spawn for the wrong curator
In a hosted game, deep copied units (Ctrl+Shift+C) spawn at the host's cursor position when pasted by any other curators (Ctrl+Shift+V) instead of their own.

Steps to reproduce:
1. Host a game.
2. Host and another player has curator.
3. Have the other player deep copy and paste a unit.
4. The unit is spawned at the hosts's cursor location.

mharis001:
- Add ability to create and place custom compositions through Zeus
- Custom compositions exist are under a special category in the empty compositions tree (marked with an icon on the right-hand side)
    - User created category and name combinations have to be unique
- Compositions can be created, edited, deleted through buttons in the panel below the empty compositions tree (like 3DEN)
    - Panel is only shown if the empty compositions tree is shown
- Placement uses vanilla Zeus placement of a helper composition which triggers spawning of the custom composition on the server
- Add `messageBox` common function to display a small confirmation box dialog
- Add `serializeObjects`/`deserializeObjects` common functions
    - Use these functions for deep copy/paste, Close #295 

<details>
<summary>Images</summary>

custom_compositions_tree

</details>


addons/common/functions/fnc_messageBox.sqf
```diff
@@ -0,0 +1,110 @@
+#include "script_component.hpp"
+/*
+ * Author: mharis001
+ * Opens are message box dialog with the given title and message text.
```
neilzar:
```suggestion
 * Opens a message box dialog with the given title and message text.
```

CreepPork:
Oh, this is going to be a slippery slope of questions coming from the majority of the end-user base:
For example, how in-depth does it save vehicles/units (e.g. fuel, weapon status, health, unit loadout)?

And there will be some who will ask to have these compositions show up in Eden or vice versa but @oOKexOo dabbled with that in Achilles and as far as I know, it wasn't possible without an elaborate extension system (and even that didn't work very well).

addons/common/functions/fnc_messageBox.sqf
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+if (canSuspend) exitWith {
+    [FUNC(messageBox), _this] call CBA_fnc_directCall;
+};
+
+if (!hasInterface) exitWith {};
+
+params [
+    ["_title", "", [""]],
```
CreepPork: Should have a default, let's say:
```suggestion
    ["_title", "Confirmation", [""]],
```

addons/compositions/stringtable.xml
```diff
@@ -0,0 +1,29 @@
...
+            <English>Name cannot be empty.</English>
+        </Key>
+        <Key ID="STR_ZEN_Compositions_CompositionAlreadyExists">
+            <English>A composition with this name and category already exists.</English>
+        </Key>
+        <Key ID="STR_ZEN_Compositions_DeleteConfirmation">
+            <English>Are you sure you want to delete this compoisition?</English>
+        </Key>
+        <Key ID="STR_ZEN_Compositions_NoObjectsSelected">
+            <English>No objects to create composition from selected</English>
```
CreepPork: This sounds really weird, needs to be rewritten.

addons/common/functions/fnc_messageBox.sqf
mharis001: I disagree, the default should be empty - to follow with some other displays like this in arma.

mharis001:
> the dropdown button should have a border or a fill

I don't think that is possible since it is a combo box.

> Could also use the given categories from Eden, like Props, etc.

I considered this but decided against it initially since, the idea behind the drop down list is for quick access to already created categories (also serves as a reference).

> Compositions when placed, do snap to the ground but they don't rotate to the ground (e.g. in Zeus moving via ALT to the ground).

Will fix by using the `surfaceNormal` command when the object is very close to the ground.

CreepPork:
The default categories should be provided as they provide quicker set up without having to create your own categories and then start adapting them to some kind of standard but teamed together with the current implementation it would work great.

Overall, the text boxes seem a tad off, I don't think their styling is in line with other text boxes.

PR #147, opened 2019-08-28
Add random placement parameter to Create Minefield module

### LINKED ISSUE #146: Randomize locations of mines using the 'Create Minefield' module
I hesitate to use the module because the created mines are so perfectly spaced, it feels very artificial. Works as a hard barrier to keep players from moving in a certain direction, but not as a hazard to clear with detectors.

**Solution you'd like:**
Using the module would remain the same, but with an extra 'random placement' toggle. In theory, it would randomize the placement of individual mines within the given radius. 3den enhanced has a similar function, and I've used it before to create natural-feeling minefields.

Can't think of an alternative way to make minefields feel more natural, aside from manually dragging individual mines around after using the module.

**Additional context:**
Just want to say you guys are doing a fantastic job.

mharis001:
- title, close #146

### DIFF
--- a/addons/modules/functions/fnc_moduleCreateMinefield.sqf
+++ b/addons/modules/functions/fnc_moduleCreateMinefield.sqf
@@ -63,7 +63,9 @@ _minesCache params ["_configNames", "_displayNames"];
 
         for "_i" from 1 to (_width / _spacing * _height / _spacing) do {
             private _position = [_area] call CBA_fnc_randPosArea;
-            _mines pushBack createVehicle [_type, _position, [], 0, "CAN_COLLIDE"];
+            private _mine = createVehicle [_type, _position, [], 0, "CAN_COLLIDE"];
+            _mine setDir random 360;
+            _mines pushBack _mine;
         };
     } else {
         private _topLeft = _position vectorAdd [-_width / 2, -_height / 2, 0];
--- a/addons/modules/functions/fnc_moduleCreateMinefield.sqf
+++ b/addons/modules/functions/fnc_moduleCreateMinefield.sqf
@@ -21,31 +21,63 @@ private _minesCache = +(uiNamespace getVariable QGVAR(minesCache));
 _minesCache params ["_configNames", "_displayNames"];
 
 [LSTRING(ModuleCreateMinefield), [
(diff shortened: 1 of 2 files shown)


CreepPork:
Are mines rotated randomly as well? Something like the APERS Mine Dispenser should be randomized to not make them deploy in the same spot.

CreepPork:
Maybe an adjustable percentage of how many mines should fill the area?

CreepPork:
And I don't know about you but the CBA random function (as in artillery too) feels not random enough (or maybe this is going to be the Spotify issue). I can see the outer ring being very popular with the algorithm and sometimes it plops mines right next to each other (as indicated in the image and when using artillery, it sometimes falls in the same place).

veteran29:
Take a look at this for uniform distibution of mines in area:
http://sqf.ovh/sqf%20math/2018/05/05/generate-a-random-position.html

mharis001:
> feels not random enough

`CBA_fnc_randPosArea` is random - uniformly distributed. Sometimes you will have mines right next to each other.

On the topic of the amount of mines, I personally think it is fine but either way, that is not the focus of this PR.

jonpas:
If it doesn't "feel" random enough then you know it's actually random. :)

addons/modules/functions/fnc_moduleCreateMinefield.sqf
```diff
@@ -21,31 +21,65 @@ private _minesCache = +(uiNamespace getVariable QGVAR(minesCache));
...
-    private _mineSpacing = -5 * _mineDensity + 30;
-    private _topLeftPos = _position vectorAdd [-_areaWidth / 2, -_areaHeight / 2, 0];
-
-    // Create minefield of given area
-    for "_i" from _mineSpacing / 2 to _areaWidth step _mineSpacing do {
-        for "_j" from _mineSpacing / 2 to _areaHeight step _mineSpacing do {
-            private _mine = createVehicle [_mineType, [0, 0, 0], [], 0, "CAN_COLLIDE"];
-            _mine setPos (_topLeftPos vectorAdd [_i, _j, 0]);
-            _createdMines pushBack _mine;
+    private _spacing = 30 + _density * -5;
```
neilzar: I would suggest the following, because the very low density feels like a medium/high density minefield.
```suggestion
    private _spacing = 60 + _density * -10;
```

PR #154, opened 2019-09-06
Attributes Rewrite

mharis001:
- Rewrite `attributes` component to work as a proper framework
    - Separation between UI handling and attribute specific functions
    - Visually the same as old attributes, for the most part
- No longer uses hard-coded dialogs for different entity types - attributes are added to "display types" and displayed based on their conditions
    - Display types "Object", "Group", "Waypoint", "Marker" are special in that they are opened when a curator entity is double-clicked
    - Every object type will have an attributes dialog now
    - Will allow for easier addition of new attributes, especially those specific to a single type of entity
- Attribute statements are only called if its corresponding control has `QGVAR(value)` defined (generally if a value was changed)
- If an object still has the old `curatorInfoType`/`curatorInfoTypeEmpty` entries defined, it will still be used (for example, modules)
- Move UI base classes (and some UI handling) to `common`
    - Add common base display classes - have the visuals of the standard Zeus displays and provide the display's config as a variable in the display namespace
    - Add common `initDisplayPositioning` function - handles adjusting the positioning of display elements based on the content height
- Cleanup and improve `inventory` component
    - No longer an attribute, but a separate display that can be opened to configure the object's inventory
- Cleanup and improve `modules` component
    - Separate from old attributes framework by using new `initDisplay` function for module dialogs
    - Dialogs can immediately delete the logic if it is not needed for the module to function without the display being closed
- Move damage components functionality (and hitpoint strings) to new `damage` component


addons/attributes/functions/fnc_bi_showCuratorAttributes.sqf
```diff
@@ -8,52 +8,52 @@
...
-    case (typeName grpNull): {
-        getText (configFile >> "CfgCurator" >> "groupInfoType");
-    };
-    case (typeName []): {
-        getText (configFile >> "CfgCurator" >> "waypointInfoType");
-    };
-    case (typeName ""): {
-        getText (configFile >> "CfgCurator" >> "markerInfoType");
+
+    private _infoTypeClass = if (isNull group _entity && {side _entity != sideLogic}) then {"curatorInfoTypeEmpty"} else {"curatorInfoType"};
```
CreepPork:
```suggestion
    private _infoTypeClass = ["curatorInfoType", "curatorInfoTypeEmpty"] select (isNull group _entity && {side _entity != sideLogic};
```

addons/attributes/script_component.hpp
```diff
@@ -24,122 +24,31 @@
...
+#define IDC_ATTRIBUTE_BACKGROUND 32
+#define IDC_ATTRIBUTE_SLIDER 33
+#define IDC_ATTRIBUTE_EDIT 34
+#define IDC_ATTRIBUTE_COMBO 35
+#define IDC_ATTRIBUTE_TOOLBOX 36
+
+#define MAX_HEIGHT POS_H(25)
+#define VERTICAL_SPACING POS_H(0.1)
+
+#define RANKS ["PRIVATE", "CORPORAL", "SERGEANT", "LIEUTENANT", "CAPTAIN", "MAJOR", "COLONEL"]
```
CreepPork: Shouldn't these be localized?

addons/attributes/functions/fnc_handleObjectPlaced.sqf
```diff
@@ -0,0 +1,30 @@
...
+ * Public: No
+ */
+
+params ["", "_object"];
+
+_object call BIS_fnc_curatorAttachObject;
+
+BIS_fnc_curatorObjectPlaced_mouseOver = curatorMouseOver;
+
+private _infoTypeClass = if (isNull group _object && {side _object != sideLogic}) then {"curatorInfoTypeEmpty"} else {"curatorInfoType"};
```
CreepPork: Select here would be better.

CreepPork:
Quite an extensive change list, raises a question about backwards compatibility but I assume there shouldn't be any issues?

addons/attributes/script_component.hpp
mharis001: There are not displayed strings.

mharis001:
Not really any backwards compatibility to preserve.

mharis001:
Decided not to implement adding attributes through config. Don't think it is needed and it is easier/cleaner to deal with the code blocks through script.

mharis001:
Not fully finished yet, still some cleanup and finalizing to do but, should be ready for general feedback.

For the most part, should be the same as the old attributes displays but feedback on things that have changed (such as, inventory attribute is always a separate display, lock is now a combo box, no rank attribute for vehicles, new damage and simulation attributes) would be good.

CreepPork:
Could you attach screenshots of the changed attributes? I personally think that the toolbox for the lock worked quite well and didn't need to be changed into a combo box which could be quite cumbersome.

addons/attributes/functions/fnc_addAttribute.sqf
```diff
@@ -0,0 +1,59 @@
+#include "script_component.hpp"
+/*
+ * Author: mharis001
+ * Adds an attribute to the given attributes display type.
+ *
+ * Arguments:
+ * 0: Display Type <STRING>
+ * 1: Label and Tooltip <STRING|ARRAY>
+ * 2: Control <STRING>
```
CreepPork: Control type string?

addons/attributes/functions/fnc_addButton.sqf
```diff
@@ -0,0 +1,31 @@
+#include "script_component.hpp"
+
```
CreepPork: Doc block

addons/attributes/functions/fnc_addTitle.sqf
```diff
@@ -0,0 +1,9 @@
+#include "script_component.hpp"
+
```
CreepPork: Doc block

addons/attributes/functions/fnc_addButton.sqf
```diff
@@ -0,0 +1,31 @@
...
+};
+
+private _buttons = GVAR(buttons) getVariable _type;
+
+if (isNil "_buttons") then {
+    _buttons = [];
+    GVAR(buttons) setVariable [_type, _buttons];
+};
+
+_buttons pushBack [_displayName, _tooltip, _statement, _condition, _closeDisplay];
```
CreepPork: Doesn't pushBack return a number value?

addons/attributes/functions/fnc_addAttribute.sqf
```diff
@@ -0,0 +1,59 @@
...
+ * 6: Condition <CODE> (default: {true})
+ * 7: Priority <NUMBER> (default: 0)
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call zen_attributes_fnc_addAttribute
+ *
+ * Public: No
```
CreepPork: Shouldn't this be public API?

addons/attributes/functions/fnc_addAttribute.sqf
```diff
@@ -0,0 +1,59 @@
...
+
+if (isNil "_attributes") then {
+    _attributes = [];
+    GVAR(attributes) setVariable [_type, _attributes];
+};
+
+_attributes pushBack [_displayName, _tooltip, _control, _valueInfo, _statement, _defaultValue, _condition, _priority];
+
+// Sort the attributes based on priority
+[_attributes, 7, false] call CBA_fnc_sortNestedArray;
```
CreepPork: Return type is incorrect.

addons/attributes/functions/fnc_confirm.sqf
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Example:
+ * [CONTROL] call zen_attributes_fnc_confirm
+ *
+ * Public: No
+ */
+
+params ["_ctrlButtonOK"];
+
+private _display = ctrlParent _ctrlButtonOK;
```
CreepPork: If you are doing this, then at least match all 3.
```suggestion
private _display  = ctrlParent _ctrlButtonOK;
```

addons/attributes/functions/fnc_gui_checkbox.sqf
```diff
@@ -0,0 +1,31 @@
+#include "script_component.hpp"
+/*
+ * Author: mharis001
+ * Initializes the checkbox attribute control type.
+ *
+ * Arguments:
+ * 0: Controls Group <CONTROL>
```
CreepPork: Pretty certain previously it was just `Control Group` + it sounds better

All files then should be updated in accordance with this suggestion and variables too.
```suggestion
 * 0: Control Group <CONTROL>
```

addons/attributes/functions/fnc_gui_code.sqf
```diff
@@ -0,0 +1,96 @@
...
+ * Example:
+ * [CONTROL, _entity, "", ["", 20, ""]] call zen_attributes_fnc_gui_code
+ *
+ * Public: No
+ */
+
+#define PREVIEW_LENGTH 30
+
+params ["_controlsGroup", "_entity", "_defaultValue", "_valueInfo"];
+_valueInfo params [["_historyVarName", "", [""]], ["_maxStatements", 20, [0]], ["_tooltip", "", [""]]];
```
CreepPork: Push one line down and make it multiple lines so it's less of a mess.

addons/attributes/functions/fnc_gui_code.sqf
```diff
@@ -0,0 +1,96 @@
...
+
+_ctrlCombo ctrlAddEventHandler ["LBSelChanged", {
+    params ["_ctrlCombo", "_index"];
+    (_ctrlCombo getVariable QGVAR(params)) params ["_controlsGroup", "_ctrlEdit", "_historyVarName"];
+
+    private _text = if (_index > 0) then {
+        private _history = profileNamespace getVariable [_historyVarName, []];
+        _history select (_index - 1)
+    } else {
+        "" // new
```
CreepPork:
```suggestion
        "" // New
```

addons/attributes/functions/fnc_gui_combo.sqf
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+params ["_controlsGroup", "_entity", "_defaultValue", "_valueInfo"];
+
+private _ctrlCombo = _controlsGroup controlsGroupCtrl IDC_ATTRIBUTE_COMBO;
+_ctrlCombo setVariable [QGVAR(params), [_controlsGroup]];
+
+{
+    _x params ["_value", "_displayName", ["_tooltip", "", [""]], ["_pictureArg", "", ["", []]]];
+    _pictureArg params [["_picture", "", [""]], ["_pictureColor", [1, 1, 1, 1], [[]], [4]]];
```
CreepPork: Push one line down

addons/attributes/script_component.hpp
```diff
@@ -2,9 +2,9 @@
 #define COMPONENT_BEAUTIFIED Attributes
 #include "\x\zen\addons\main\script_mod.hpp"
 
-// #define DEBUG_MODE_FULL
-// #define DISABLE_COMPILE_CACHE
-// #define ENABLE_PERFORMANCE_COUNTERS
+#define DEBUG_MODE_FULL
+#define DISABLE_COMPILE_CACHE
+#define ENABLE_PERFORMANCE_COUNTERS
```
CreepPork: Should be disabled.

addons/inventory/script_component.hpp
```diff
@@ -2,9 +2,9 @@
 #define COMPONENT_BEAUTIFIED Inventory
 #include "\x\zen\addons\main\script_mod.hpp"
 
-// #define DEBUG_MODE_FULL
-// #define DISABLE_COMPILE_CACHE
-// #define ENABLE_PERFORMANCE_COUNTERS
+#define DEBUG_MODE_FULL
+#define DISABLE_COMPILE_CACHE
+#define ENABLE_PERFORMANCE_COUNTERS
```
CreepPork: Should be disabled

addons/attributes/functions/fnc_gui_combo.sqf
neilzar: I wouldn't say so, both are params and it is cleaner to keep those together

addons/attributes/functions/fnc_gui_checkbox.sqf
mharis001: No, it is supposed to be "controls group".

addons/attributes/functions/fnc_addAttribute.sqf
mharis001: Not intended to be.

addons/attributes/script_component.hpp
mharis001: Not finished yet. Only ready for feedback on general functionality, code still needs some work/cleanup.

mharis001:
> Every killed unit will only have the execute attribute, they should have the name field at least added to that.

Dead units do not have a the name attribute because shortly after they die `setName` does not work and `name` returns "Error: No unit".

> And how well does it work with the ability of mission makers hiding attributes and attribute buttons? Something like with BIS_fnc_setCuratorAttributes?

Objects using BI's framework (`curatorInfoType`/`curatorInfoTypeEmpty`) are still fully supported. Support in ZEN's attributes displays will stay the same as current master.

addons/attributes/script_component.hpp
CreepPork: Yeah, if it was a draft PR then you could leave it enabled but the it's no longer a draft.

CreepPork:
Didn't dead units already have a name field in vanilla? And if it is so problematic then you could get an event when the unit dies and just add it into a variable for the unit (not the best idea, I'd admit)?

neilzar:
I checked again, shortly after death you can still set and get the name. However, after more than a minute you get the `no unit` error. So I guess there is no issue with not having the name field.

addons/attributes/functions/fnc_addButton.sqf
mharis001: Yes, it does.

mharis001:
I think this should (finally) be good to go.

Combined display data into one namespace and added requirement of registering a display with `addDisplay` before  attributes or buttons can be added.

addons/attributes/functions/fnc_gui_icons.sqf
```diff
@@ -0,0 +1,104 @@
...
+    } forEach _iconControls;
+
+    _controlsGroup setVariable [QGVAR(value), _selectedIndex];
+};
+
+private _display = ctrlParent _controlsGroup;
+private _iconControls = [];
+
+{
+    _x params ["_icon", "_tooltip", "_posX", "_posY", "_size", ["_color", [1, 1, 1], [[]], [3, 4]], ["_condition", {true}, [{}]]];
```
neilzar: Would help to split the array over newlines, it is quite long.

addons/attributes/functions/fnc_gui_waypoint.sqf
```diff
@@ -0,0 +1,52 @@
...
+    };
+} forEach _waypointTypes;
+
+_ctrlToolbox ctrlAddEventHandler ["ToolBoxSelChanged", {
+    params ["_ctrlToolbox", "_index"];
+    (_ctrlToolbox getVariable QGVAR(params)) params ["_controlsGroup"];
+
+    private _value = _ctrlToolbox getVariable str _index;
+    _controlsGroup setVariable [QGVAR(value), _value];
+}];
```
neilzar: Returns Number but header says None

addons/attributes/functions/fnc_gui_code.sqf
```diff
@@ -0,0 +1,104 @@
...
+private _ctrlEdit  = _controlsGroup controlsGroupCtrl IDC_ATTRIBUTE_EDIT;
+
+_ctrlCombo setVariable [QGVAR(params), [_controlsGroup, _ctrlEdit, _historyVarName]];
+_ctrlCombo lbAdd localize "STR_A3_RscAttributeTaskDescription_New";
+_ctrlCombo lbSetCurSel 0;
+
+{
+    // Shorten long statements and add "..." to the end
+    if (count _x > PREVIEW_LENGTH) then {
+        _x = format ["%1%2", _x select [0, PREVIEW_LENGTH], toString [ASCII_PERIOD, ASCII_PERIOD, ASCII_PERIOD]];
```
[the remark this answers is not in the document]
mharis001: I tried it, I think with the font three periods look a bit better than the ellipsis.

addons/attributes/functions/fnc_addDisplay.sqf
```diff
@@ -0,0 +1,35 @@
+#include "script_component.hpp"
+/*
+ * Author: mharis001
+ * Registers the given attributes display type allowing attributes and buttons to be added to it.
+ * The title is formatted with an entity specific description as "%1".
```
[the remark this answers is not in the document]
mharis001: Means that the "entity specific description" will be "%1" when using the `format` command on the title.

addons/attributes/functions/fnc_gui_checkbox.sqf
```diff
@@ -0,0 +1,31 @@
...
+ * Public: No
+ */
+
+params ["_controlsGroup", "_entity", "_defaultValue"];
+
+private _ctrlCheckBox = _controlsGroup controlsGroupCtrl IDC_ATTRIBUTE_CHECKBOX;
+_ctrlCheckBox setVariable [QGVAR(params), [_controlsGroup]];
+_ctrlCheckbox cbSetChecked _defaultValue;
+
+_ctrlCheckbox ctrlAddEventHandler ["CheckedChanged", {
```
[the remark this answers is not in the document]
mharis001: This is a private UI function - "None", at least in ZEN, has been treated as `nil` or "has no meaning".

mharis001:
I think the combo looks cleaner in the UI and prefer it with the icons. Don't see why showing "Error: No Unit" is a good thing for the name field.

CreepPork:
Error: No Unit doesn't appear all the time and does show the name, as @neilzar said. Only after some time, it transitions into that error message.

mharis001:
That transition happens shortly after, and could happen while display is open. I think it is better to completely avoid the situation where changing the name has no effect.

CreepPork:
You don't have to change the name, for missions, it is useful to see which particular unit had been killed as there might be info you as the mission creator added and it can help to identify the specific unit. Plus, ACE dog tags go around this issue and could be investigated to use the same approach as ACE does.

neilzar:
I think it is not worth to show the name for dead units, the name can only be gotten and set in the minute after death. After that it is useless. Won't help for identification either, because it gives the error.

As for the lock, I think it is better to have the drop-down menu because the symbols won't be quick enough to convey the meaning of that option. It is better to have the text immediately visible without having it as a tooltip, especially for new Zeuses.

CreepPork:
I'm not thinking about how it is in vanilla but a toolbox control. I think that would allow all 4 options to be visible but being easily accessible and not hidden away and you can add the tooltip as well.

neilzar:
A toolbox could work. The only downside is that it can't show the icons, but I'm not sure how necessary it is to have those.

mharis001:
Going to keep the lock attribute as a combo box. There is no significant downgrade is accessibility and it is visually superior.

neilzar:
I think that a toolbox looks nicer, less empty area in most cases but still the same visual space taken up. Obviously there is a limit to the length of the options, but in this case, they will fall within those limits.

addons/attributes/functions/fnc_gui_waypoint.sqf
```diff
@@ -0,0 +1,52 @@
...
+private _waypointScript = waypointScript _entity;
+private _waypointTypes  = uiNamespace getVariable QGVAR(waypointTypes);
+
+{
+    _x params ["_name", "_type", "_script"];
+
+    private _index = _ctrlToolbox lbAdd _name;
+    _ctrlToolbox setVariable [str _index, [_type, _script]];
+
+    if (_type == _waypointType && {_type != "SCRIPTED" || {_script == _waypointScript}}) then {
```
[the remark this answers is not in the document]
mharis001: Should be case-insensitive.

addons/attributes/functions/fnc_bi_showCuratorAttributes.sqf
```diff
@@ -1,59 +1,60 @@
...
+    if (isClass (configFile >> _infoType)) then {
+        private _curator = getAssignedCuratorLogic player;
+        private _attributes = [_curator, _entity] call BIS_fnc_curatorAttributes;
+
+        if (_attributes isEqualTo []) then {
+            false breakOut "Main";
+        } else {
+            BIS_fnc_initCuratorAttributes_target = _entity;
+            BIS_fnc_initCuratorAttributes_attributes = _attributes;
+            createDialog _infoType breakOut "Main";
```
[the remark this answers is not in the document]
mharis001: Unnecessary, and putting parentheses around unary commands is generally avoided in the project.

addons/attributes/functions/fnc_addDisplay.sqf
```diff
@@ -0,0 +1,35 @@
+#include "script_component.hpp"
+/*
+ * Author: mharis001
+ * Registers the given attribute display type allowing attributes and buttons to be added to it.
+ * The title is formatted with an entity description as the "%1" format specifier.
+ * If the title is empty, a generic "Edit %1" title will be used.
+ *
+ * Arguments:
+ * 0: Display Type <STRING>
+ * 1: Title <STRING> (default: "")
```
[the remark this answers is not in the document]
mharis001: The description does not describe the default, it describes the effect of an empty title `""`. A bit weird but, I think the it match the default used in `params` is fine.

addons/attributes/initAttributes.sqf
```diff
@@ -0,0 +1,955 @@
+// - Object -------------------------------------------------------------------
+
+["Object", "", true] call FUNC(addDisplay);
```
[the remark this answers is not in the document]
mharis001: I'd rather keep these in one file, everything with adding the default attributes being contained in this one file makes sense and is better than the component's folder from having 8 different files for this. I don't think this is messy, just long.

addons/attributes/initAttributes.sqf
```diff
@@ -0,0 +1,955 @@
+// - Object -------------------------------------------------------------------
```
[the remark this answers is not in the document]
mharis001: Don't think `#region` exists for SQF, or that I've never seen it used.

addons/attributes/functions/fnc_bi_showCuratorAttributes.sqf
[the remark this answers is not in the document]
CreepPork: It is, but in this case I would advocate it because it is far more readable to someone who sees it for the first time.

addons/attributes/initAttributes.sqf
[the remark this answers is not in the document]
mharis001: Dashes are just for some visual separation, going to keep for now.

addons/attributes/functions/fnc_gui_code.sqf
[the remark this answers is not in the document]
mharis001: I don't think that is needed.

mharis001:
For now, I think it best to keep this private. Can be made a public API later.

mharis001:
Can look at further cleanup and improvements in a follow up PR.

PR #158, opened 2019-09-08
Check surface intersections when determining position from screen

mharis001:
- Add `zen_common_fnc_getPosFromScreen` - function to determine world position based on screen position which checks any intersections with surfaces
- Use this new function in `zen_common_fnc_getTargetPos` and context menu's `_contextPosASL`
- Will allow, for example, teleport context menu actions to allow placing a unit on top of a building (not always the ground)


addons/common/functions/fnc_getPosFromScreen.sqf
```diff
@@ -0,0 +1,40 @@
+#include "script_component.hpp"
+/*
+ * Author: mharis001
+ * Returns a world position based on the given screen position.
+ * Will attempt to find a flat position on an intersecting surface if possible.
+ *
+ * Arguments:
+ * 0: Screen Position <ARRAY> (default: getMousePosition)
```
CreepPork: Does it really need to be wrapped into another array?

mharis001: Yes, cleaner `params`.

PR #162, opened 2019-09-13
Add Toggle Flashlights module

mharis001:
- Add module to toggle (and add) flashlights to infantry units


CreepPork:
If there is a flashlight module then there should be one for laser pointers and NVGs.

mharis001:
This will work for any weapons that have any compatible flashlight attachments.

PR #163, opened 2019-09-14
Add icons to heal children actions, Move player action up

mharis001:
- Add the medical cross icon to heal action children
- Move the player child action up
- Move main action condition to the All action
    - Actions in context menu don't show if they have no children and an empty statement
- Use vanilla string for "AI"


mharis001:
@3Mydlo3

PR #168, opened 2019-09-16
Add context menu actions to copy/paste vehicle appearance

### LINKED ISSUE #166: Copy/Paste Vehicle Appearance
It takes time to setup multiple vehicles with the same appearance.

**Solution you'd like:**
Ideally it'd be nice if Zen could apply it on Copy/Paste in Zeus, same with loadouts too.
If that isn't possible at the moment add Copy/Paste to the context menu like there currently is for Edit Loadout would be nice.

mharis001:
- title, Close #166


CreepPork:
I'd personally prefer a keybind for this. Maybe a context menu option is a bit overkill but I'd still leave it be. Something like a customizable version of Achilles' deep copy and paste (vehicle style, loadouts, same applies to units) could be integrated with it.

PR #172, opened 2019-09-21
Keep factions visible when clicking the collapse all button

mharis001:
- title, fix annoying behaviour resulting from having to expand the useless "side" node of group trees to view available factions
- technically, the group trees are not fully collapsed after this change but I think most users will prefer this behaviour


CreepPork:
Is it possible to just remove the top node, or does it require an entry point node? Oh, and there might be some compatibility issues when doing that.

mharis001:
I originally considered doing that but decided against it due to the overhead it would create (+5 trees to fill on load) for a minor QOL improvement.

Issue #165, opened 2019-09-15
Ambient FlyBy Enter Command


GeneralKong:
**The Request:**
Place-able arrows for ambient flyby module

**Details:**
The Ability to place down the Ambient Flyby in a way to allow you to place multiple arrows (like you get when you are placing down the USS Freedom and USS Liberty), and when you press enter, it will spawn them all at the same time in the direction the are pointing, making it easier to have them fly in formation

callofmarty:
Hi, not sure if i can post it here. Is it possible to make the flyby module remember last entered settings?

mharis001:
This seems very cumbersome to try to implement and I am not sure if its the best UX as well. The Ambient Flyby module recently received some improvements such as selection saving and an amount of aircraft option.

I think the combination of those two should make what you want to accomplish much easier (and remove the need to add any sort of ENTER confirmation to the module).

GeneralKong:
Agreed, with the new changes there really is no need anymore for this change

Issue #102, opened 2019-08-05
Update Editable objects module doesn't cache previously used values


CreepPork:
https://gyazo.com/d4c478981b7bd899d1ddc80989ade71b

mharis001:
Not a bug, selection saving isn't implemented for the editable objects display.

PR #174, opened 2019-09-22
Add selection saving to Ambient Flyby dialog

mharis001:
- title, https://github.com/zen-mod/ZEN/issues/165#issuecomment-533880549


CreepPork:
It should be noted that other dialogs also lack saving functionality and are a pain, e.g. #102.

PR #179, opened 2019-09-27
Add fixed altitude option to Ambient FlyBy

neilzar:
- title, aircraft will stay at the same altitude rather than follow the terrain when the option is selected.


addons/modules/RscAttributes.hpp
```diff
@@ -808,38 +808,56 @@ BEGIN_MODULE_DIALOG(RscAmbientFlyby)
                 w = POS_W(2.4);
                 h = POS_H(1);
             };
+            class FixedLabel: EGVAR(attributes,RscLabel) {
```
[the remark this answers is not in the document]
neilzar: That is actually really elegant.

addons/modules/RscAttributes.hpp
```diff
@@ -793,53 +793,67 @@ BEGIN_MODULE_DIALOG(RscAmbientFlyby)
                 text = ECSTRING(common,Height);
                 tooltip = CSTRING(ModuleAmbientFlyby_Height_Tooltip);
                 y = POS_H(5.6);
+                h = POS_H(2.1);
+            };
+            class Fixed: Direction {
```
mharis001: I think `class HeightMode` would be a better name.

addons/modules/stringtable.xml
```diff
@@ -223,6 +223,18 @@
...
+            <English>Follow Terrain</English>
+        </Key>
+        <Key ID="STR_ZEN_Modules_ModuleAmbientFlyby_Fixed_ASL">
+            <English>Fixed Altitude</English>
+        </Key>
+        <Key ID="STR_ZEN_Modules_ModuleAmbientFlyby_Fixed_AGL_Tooltip">
+            <English>The aircraft will follow the terrain, staying at the same height relative to the ground. Based on AGL.</English>
+        </Key>
+        <Key ID="STR_ZEN_Modules_ModuleAmbientFlyby_Fixed_ASL_Tooltip">
+            <English>The aircraft will fly at a constant altitude, only following the terrain if less than 10 meters from the ground.\nBased on ASL, final altitude is the elevation of the module plus the height given.</English>
```
mharis001: For some reason, `\n` does not work for toolbox option tooltips. Should revert.

addons/modules/functions/fnc_moduleAmbientFlyby.sqf
```diff
@@ -23,11 +23,15 @@
 
 #define SPAWN_OFFSET -30
 
-params ["_aircraftType", "_position", "_height", "_distance", "_direction", "_speed", "_amount"];
+params ["_aircraftType", "_position", "_height", "_useASL", "_distance", "_direction", "_speed", "_amount"];
```
mharis001: Add to the header.

addons/modules/functions/fnc_gui_ambientFlyby.sqf
```diff
@@ -102,6 +101,9 @@ private _ctrlHeightSlider = _display displayCtrl IDC_AMBIENTFLYBY_HEIGHT_SLIDER;
 private _ctrlHeightEdit   = _display displayCtrl IDC_AMBIENTFLYBY_HEIGHT_EDIT;
 [_ctrlHeightSlider, _ctrlHeightEdit, 10, 5000, _height, 50] call EFUNC(common,initSliderEdit);
 
+private _ctrlFixed = _display displayCtrl IDC_AMBIENTFLYBY_HEIGHT_MODE;
+_ctrlFixed lbSetCurSel _useASL;
```
mharis001: `_ctrlHeightMode` and move above the slider stuff (L100).

addons/modules/functions/fnc_gui_ambientFlyby.sqf
```diff
@@ -153,11 +158,11 @@ private _fnc_onConfirm = {
     private _ctrlAmount = _display displayCtrl IDC_AMBIENTFLYBY_AMOUNT;
     private _amount = lbCurSel _ctrlAmount;
 
-    GVAR(lastAmbientFlyby) = [_side, _faction, _aircraft, _direction, _height, _distance, _speed, _amount];
+    GVAR(lastAmbientFlyby) = [_side, _faction, _aircraft, _direction, _height, _useASL, _distance, _speed, _amount];
 
     [
         QGVAR(moduleAmbientFlyby),
-        [_aircraftType, ASLtoAGL getPosASL _logic, _height, _distance, _direction, _speed, _amount + 1]
+        [_aircraftType, getPosASL _logic, _height, _useASL == 1, _distance, _direction, _speed, _amount + 1]
```
mharis001: Trivial but, I think it would be better to keep the ordering so it follows the UI.

mharis001:
Parameter `_useASL` needs to be added to the header.

PR #183, opened 2019-09-30
Move garage GETDLC macro to common function

mharis001:
- title


addons/common/functions/fnc_getDLC.sqf
```diff
@@ -0,0 +1,31 @@
+#include "script_component.hpp"
+/*
+ * Author: mharis001
+ * Returns the DLC that the given config belongs to.
+ *
+ * Arguments:
+ * 0: Config <CONFIG>
+ *
+ * Return Value:
+ * DLC <STRING>
```
CreepPork: DLC what? Name? Classname?

PR #191, opened 2019-10-02
Add translations CI workflow

mharis001:
- Add CI workflow to update issue with translation status
- Based off acemod/ACE3#7208

### DIFF
--- a/.github/workflows/translations.yml
+++ b/.github/workflows/translations.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
     - master
-    - translations-ci
 
 jobs:
   update:
--- a/.github/workflows/translations.yml
+++ b/.github/workflows/translations.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
     - master
+    - translations-ci
 
 jobs:
   update:
--- a/.github/workflows/translations.yml
+++ b/.github/workflows/translations.yml
@@ -1,0 +1,24 @@
+name: Translations
+
+on:
+  push:
+    branches:
+    - master
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout the source code
+      uses: actions/checkout@master
+    - name: Install Python packages
+      run: |
+        pip3 install wheel
+        pip3 install setuptools
+        pip3 install pygithub
+        pip3 install pygithub3
+    - name: Deploy
+      if: github.repository == 'zen-mod/ZEN' && ! contains(github.event.head_commit.message, '[ci skip]')
+      env:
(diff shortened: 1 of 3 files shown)


CreepPork:
I don’t think this should using GitHub Actions. We already are using Pipelines which underneath is basically the same thing.

jonpas:
Pipelines are what? GitHub Actions for this is perfect, you already get secret user provided and is integrated with GitHub.

CreepPork:
Azure Pipelines. We use it for the build system. This way you add another CI that takes up rate limits too. Plus, it'd be faster for it to be just integrated with Pipelines. The structure is near identical.

jonpas:
> takes up rate limits

What rate limits?

> because, for example, it saves checkout times

Either you have multiple checkouts for parallel work, or you have 1 checkout and serial work. Former is faster almost always.

---

It makes way more sense to have this on Actions with integrated secrets and GitHub interoperability. However, I am not defending either, merely questioning your reasons which I do not understand.

.github/workflows/translations.yml
```diff
@@ -0,0 +1,24 @@
...
+    - name: Checkout the source code
+      uses: actions/checkout@master
+    - name: Install Python packages
+      run: |
+        pip3 install wheel
+        pip3 install setuptools
+        pip3 install pygithub
+        pip3 install pygithub3
+    - name: Deploy
+      if: github.repository == 'zen-mod/ZEN' && ! contains(github.event.head_commit.message, '[ci skip]')
```
CreepPork: Is `ci skip` even necessary at any point? For some other builds ok, but for this? And as it's running only on master and master is only updated via PRs, then added with squash merging, the PR should start with `ci skip`.

jonpas: No, this is used in ACE3 for documentation updates via CI. There is no reason to not include it here though, currently Actions do not support `ci skip` natively.

CreepPork: CI skip just doesn't go into this project and can be safely removed and by any chance, a different doc system will be chosen, it will damn sure not push commits to master.

mharis001:
I think GitHub Actions is the way to go with this. Can move the azure pipelines build to actions at some point too, if we want it all on one.

bux:
Both is Microsoft sooooo.... ¯\\_(ツ)_/¯

jonpas:
_One is so nicely integrated sooooo.... ¯\(ツ)/¯_

This is not a discussion for here. :P

.github/workflows/translations.yml
mharis001: There is no harm in leaving the "ci skip" check in.

PR #192, opened 2019-10-03
Add coding guidelines documentation

mharis001:
- Add coding guidelines documentation
- For the most part same as ACE3's, with some tweaks

### DIFF
--- a/docs/_sidebar.md
+++ b/docs/_sidebar.md
@@ -9,3 +9,5 @@
   - [<i class="fas fa-cog"></i> Custom Modules](/custom_modules.md)
   - [<i class="far fa-window-restore"></i> Dynamic Dialog](/dynamic_dialog.md)
   - [<i class="fas fa-broadcast-tower"></i> Public Events](/public_events.md)
+- **Development**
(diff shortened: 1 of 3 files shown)


CreepPork:
I’m not sure ZEN needs it’s own page for coding guidelines. It’s pretty clear that ZEN uses ACE’s and then you have to maintain another page for any changes. Plus it’s more difficult for other contributors to be compliant due to minute differences that could be present between the two docs.

mharis001:
This isn't something that is going to need to be updated everyday. I don't see how it is more difficult for anyone to contribute by also providing the CBA/ACE style guidelines on ZEN's documentation website as well.

CreepPork:
It doesn't hurt for users when the docs match. But down the line, you'll have discrepancies between the two docs as they change. Then which one becomes the source of truth?
Plus, why is it even needed here? It's pretty obvious that ZEN follows ACE's conventions so why can't people just look at the ACE docs?

mharis001:
Well, ZEN's would be the one to use. I really don't see why ZEN having its own coding guidelines is a problem, even if they are similar to ACE's.

PR #193, opened 2019-10-04
Add garage button to apply customization to all selected vehicles

### LINKED ISSUE #97: Garage doesn't apply selection to all selected vehicles but only the first one
https://gyazo.com/a6108267d250d081f34219b5d32b60a6

mharis001:
- title, close #97

### DIFF
--- a/addons/garage/functions/fnc_applyToAll.sqf
+++ b/addons/garage/functions/fnc_applyToAll.sqf
@@ -1,0 +1,27 @@
+#include "script_component.hpp"
+/*
+ * Author: mharis001
+ * Applies the customization of the current garage vehicle
+ * to all selected vehicles of the same type.
+ *
+ * Arguments:
+ * None
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call zen_garage_fnc_applyToAll
+ *
+ * Public: No
+ */
+
+(GVAR(center) call BIS_fnc_getVehicleCustomization) params ["_texture", "_animations"];
+
+private _vehicleType = typeOf GVAR(center);
+
+{
+    if (typeOf _x isEqualTo _vehicleType) then {
+        [_x, _texture, _animations, true] call BIS_fnc_initVehicle;
+    };
+} forEach SELECTED_OBJECTS;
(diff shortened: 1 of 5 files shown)


addons/garage/functions/fnc_applyToAll.sqf
```diff
@@ -0,0 +1,27 @@
...
+
+(GVAR(center) call BIS_fnc_getVehicleCustomization) params ["_texture", "_animations"];
+
+private _vehicleType = typeOf GVAR(center);
+
+{
+    if (typeOf _x isEqualTo _vehicleType) then {
+        [_x, _texture, _animations, true] call BIS_fnc_initVehicle;
+    };
+} forEach SELECTED_OBJECTS;
```
neilzar: I believe in case the last object in `SELECTED_OBJECTS` is the same type, so the if statement is true, the forEach will return a boolean because that is the return of `BIS_fnc_initVehicle`. Return value says `None`.

mharis001: This is not a problem.

neilzar: It would be nice to either have a nil at the end to make sure it always returns nothing, or change the header to say it can return a boolean.

mharis001: No need, I've always used "None" as nothing AKA `nil` or no meaning in the project.

jonpas: Basically think of None return also as "not to be used". :D

CreepPork: How is it not a problem? The doc block indicates that it doesn't return anything but in fact, it does sometimes return a value and that should be fixed by adding a `nil` at the end of the file. This could bite you in the ass later if you don't know that it could return such a value half of the time.

mharis001: What happens when a function calls several other functions (each of which call other functions) that could all be possible returns, which may conditionally return different value types.

When a function's return value is meaningful, its stated in the header. Otherwise, its "None".

jonpas: If you are writing code that does something with a function without a meaningful return value and you rely on that, you are already making a huge mistake. If your code fails because a function doesn't specifically return `nil` then why not pass your own proper `nil` which you can be sure will always work in the future, even if that function gets a meaningful return value in the future (which would also break if you would rely on that btw).

And changing return values of functions is usually not as backwards incompatible when used properly.

PR #194, opened 2019-10-04
Allow toggling garage interface using mouse buttons

mharis001:
- Add ability to toggle garage interface by right clicking and showing hidden interface by left clicking


CreepPork:
Documentation for users should be added, otherwise, they won't have a clue that this feature exists.

mharis001:
This is the same behaviour as BI's garage - documentation for the garage, in general, should be done in another PR.

PR #199, opened 2019-10-05
Add fastrope waypoint

mharis001:
- Add scripted fastrope waypoint (requires `ace_fastroping`)

### DIFF
--- a/addons/ai/functions/fnc_waypointFastrope.sqf
+++ b/addons/ai/functions/fnc_waypointFastrope.sqf
@@ -52,6 +52,10 @@ private _skill = skill _driver;
 _driver allowFleeing 0;
 _driver setSkill 1;
 
+// Set the group's behaviour to careless to prevent it from flying away in combat
+private _behaviour = behaviour _vehicle;
+_group setBehaviour "CARELESS";
+
 // Set the helicopter to fly at the fastrope height
 _vehicle flyInHeight FASTROPE_HEIGHT;
 
@@ -114,5 +118,6 @@ waitUntil {  _vehicle getVariable ["ace_fastroping_deployedRopes", []] isEqualTo
 _vehicle call ace_fastroping_fnc_stowFRIES;
 
 _driver setSkill _skill;
+_group setBehaviour _behaviour;
 
 true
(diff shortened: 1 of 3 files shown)


CreepPork:
Support for `Advanced Rappelling` would be very recommended. On Achilles, we ran into issues that on multiplayer ACE's rappelling was quite unstable and preferred AR over ACE's wherever possible.

jonpas:
Just fyi, Rappelling != Fast-Roping. 

Have you ever taken a look at AR code? Have fun with that! :D

But that is for another issue/PR, it is a completely different thing. 

---

Fastroping works fine, observe proper procedures and not just mash "fastrope", it is not meant for that, if everyone jumps out at the same time IRL you'll get a similar end. :D

CreepPork:
Could you enlighten me on the _major_ differences? AR just adds the ability to stop while on the rope. The ropes deploying from the helo don't matter.

So what about the code? You're not porting their code into ZEN, you just need to check that the deploy function is a publicVariable'd function and call it via `[helo] call AR_Rappel_All_Cargo`.

A major benefit is that it's server-side only.

I haven't played around with fastrope in a while but afaik it hasn't changed much. It had general instabilities in multiplayer and if I'm not mistaken with non-local units. Maybe @oOKexOo can give a better description as he was the one working on the fast-roping feature more.

jonpas:
I am just telling you it's not the same thing, you called ACE's "rappelling", while that is going down a building, rock-face etc.

And I am just saying have fun looking at it, as you already did. The rest is about game play, not interfacing with it as I see it was clearly misused, thus seperated with a horizontal line.

It is a thing for another issue/PR, I don't see what it has to do with this one, it is entirely different handling.

CreepPork:
No, you're getting it muddled up with Advanced Urban Rappelling by duda. Advanced Rappelling is used from helicopters while the urban one is from buildings, rocks and other stuff as you mentioned.

jonpas:
No, just this:

> ACE's rappelling

That mod is then named wrong though, thanks for telling me. :D Both mods are pretty much the same, same author, same single-file code, so ZEN could support both if wanted.

But anyways, this PR looks good to me.

jonpas:
Issue you linked, helicopter is too high, you have to get it lower. AI pilot increases height if there is stuff below it. And if that's not the case, open an issue at ACE3, because there were no reports where that wasn't the case so far.

mharis001:
ACE3 is the only mod ZEN provides direct support for.

PR #204, opened 2019-10-06
Adjusted wording for attribute strings

CreepPork:
- Change the English strings for two keys (very minor changes)

Should the key IDs be updated as well?


neilzar:
No need to update the keys.

PR #211, opened 2019-10-13
Add Spawn Reinforcements module

### LINKED ISSUE #4: Feature Request: Reinforcements
In the vein of Achilles, a module that sees use fairly frequently by me is the Reinforcement module, allowing the selection of 

- The team providing the vehicle

- The vehicle in question (And number of non-crew slots)

- If the vehicle will remain at the drop point, or return from where it was summoned and despawn

- Insertion Method (if from Helicopter/Plane, ie landing or paradropping or rappelling (via ACE or Adv. rappelling)

- 'Fill' team (BLU/IND/OP)

- Unit fill. In Achilles, this is a maximum of 1 squad, chosen from Compositions of the team chosen.

- Deployed unit behaviour (Safe/Aware/Combat)

This specifically required 3 modules in Achilles- LZ (Allowing multiple, and given individual names- This is where the vehicle lands/stops/etc, deploying units), RP (where said deployed units were told to move), and the actual Spawn Units module, which created the vehicle at the module location and gave it the tasking to perform said troop dropoff.

Personally I would like to see the option of non-composition selection, or more than 1 squad (as squad sizes are not always equivalent to transportation capability). This could also be the framework for a 'supply drop' module, another feature Achilles had, allowing AI drop of vehicles/containers, etc.

mharis001:
- Add "Spawn Reinforcements" module
- Add "Create LZ" and "Create RP" modules
- Add `position_logics` component - unified handling for position modules
    - Provides helper functions for dealing with position logics
- Change "Create Target" module to use position logics system
- Add `getCargoPositionsCount` common function
- Add `getVehicleIcon` common function
- Close #4


neilzar:
A couple of things I would like to see as a feature:
- I would like the ability to select one of the premade groups rather than having to custom make it. Could be achieved by adding a toolbox that has the options `group creator` and `premade groups`, or something along those lines. When the premade groups option is selected the right side is disabled and the select list is filled with groups that can fit into the selected vehicle rather than all units. Then the selected group in the left list is the group that will spawn.
- Currently when the helicopter is set to `Stay at LZ` they will fly off and stay in a random area of the map after para dropping the group. I haven't tested the stay option with landing, but I would like to see the helicopter return to the LZ with a `Hold` waypoint, or possibly even a `Search and Destroy`. I personally use the latter a lot because the helicopters will provide great air support and can be insanely deadly when the LZ is near an open field. It could also be another combo-/toolbox when selecting `Stay at LZ`, the option for `Hold` or `Search and Destroy`.

addons/modules/functions/fnc_compileReinforcements.sqf
```diff
@@ -0,0 +1,75 @@
...
+        };
+
+        _categoryList pushBack _className;
+    };
+};
+
+private _vehicles = [];
+private _infantry = [];
+
+
```
CreepPork: Double new line.

addons/modules/script_component.hpp
```diff
@@ -2,9 +2,9 @@
 #define COMPONENT_BEAUTIFIED Modules
 #include "\x\zen\addons\main\script_mod.hpp"
 
-// #define DEBUG_MODE_FULL
-// #define DISABLE_COMPILE_CACHE
-// #define ENABLE_PERFORMANCE_COUNTERS
+#define DEBUG_MODE_FULL
+#define DISABLE_COMPILE_CACHE
+#define ENABLE_PERFORMANCE_COUNTERS
```
CreepPork: Should be disabled.

addons/modules/functions/fnc_moduleSpawnReinforcements.sqf
```diff
@@ -0,0 +1,134 @@
...
+// Use normal transport unload if fastroping is not available
+if (_insertionMethod == 2 && {!isClass (configFile >> "CfgPatches" >> "ace_fastroping") || {getNumber (_vehicleConfig >> "ace_fastroping_enabled") == 0}}) then {
+    _insertionMethod = 0;
+};
+
+// Add waypoint to make vehicle drop units off at the LZ
+// Handle insertion method parameter for air vehicles
+private _waypoint = _vehicleGroup addWaypoint [_positionLZ, WAYPOINT_RADIUS];
+
+if (_isAir && {_insertionMethod > 0}) then {
```
neilzar: Helicopter will land when paradrop is selected.

neilzar:
Helis won't do anything most of the time when set to stay at LZ, they just fly in circles because they are in careless.

addons/common/script_component.hpp
```diff
@@ -2,9 +2,9 @@
 #define COMPONENT_BEAUTIFIED Common
 #include "\x\zen\addons\main\script_mod.hpp"
 
-// #define DEBUG_MODE_FULL
-// #define DISABLE_COMPILE_CACHE
-// #define ENABLE_PERFORMANCE_COUNTERS
+#define DEBUG_MODE_FULL
+#define DISABLE_COMPILE_CACHE
+#define ENABLE_PERFORMANCE_COUNTERS
```
CreepPork: Should be disabled.

addons/modules/CfgFactionClasses.hpp
```diff
@@ -34,6 +34,11 @@ class CfgFactionClasses {
         priority = 2;
         side = 7;
     };
+    class GVAR(Reinforcements) {
+        displayName = "Reinforcements";
```
CreepPork: Localized.

addons/position_logics/functions/fnc_select.sqf
```diff
@@ -0,0 +1,56 @@
...
+if (_list isEqualTo []) exitWith {objNull};
+
+// Convert mode string to mode number
+if (_mode isEqualType "") then {
+    _mode = (["random", "nearest", "farthest"] find _mode) - 3;
+};
+
+// Select position logic based on the mode
+switch (_mode) do {
+    case -3: { // Random
```
CreepPork: I'd personally, prefer them to be pushed one line up or down, not to be on the same line.

neilzar:
> From the Current Group column, the user should be able to remove units via double-click or right-click, the delete key isn't enough.
If we go with this I'd go for the single right-click, it seems more intuitive to me.

CreepPork:
Single-click should be for highlighting only.

addons/position_logics/functions/fnc_select.sqf
mharis001: Going to leave like this to match rest of the project.

mharis001:
Don't single right-click is possible to do, only other event is `onLBDblClick`.

The indent for the categories isn't perfect but it is there because having no ident looks worse as the text is not aligned with the other combo boxes.

addons/modules/functions/fnc_moduleAmbientFlyby.sqf
```diff
@@ -99,7 +104,7 @@ _waypoint setWaypointCombatMode "BLUE";
 _waypoint setWaypointSpeed _speed;
 
 // Delete aircraft, crew, and group once end waypoint is reached
-_waypoint setWaypointStatements ["true", "private _group = group this; private _aircrafts = []; {if !(vehicle _x in _aircrafts) then {_aircrafts pushback vehicle _x}; deleteVehicle _x} forEach thisList; {deleteVehicle _x;} forEach _aircrafts; deleteGroup _group"];
+_waypoint setWaypointStatements ["true", "private _group = group this; private _aircrafts = []; {_aircrafts pushBackUnique vehicle _x; deleteVehicle _x} forEach thisList; {deleteVehicle _x} forEach _aircrafts; deleteGroup _group"];
```
shukari: The warning text of deleteVehicleCrew (https://community.bistudio.com/wiki/deleteVehicleCrew) seams interesting for this statement.

addons/modules/functions/fnc_moduleSpawnReinforcements.sqf
```diff
@@ -0,0 +1,136 @@
...
+// Increase the skill of the driver/pilot for better driving/flying
+driver _vehicle setSkill 1;
+
+// Spawn the infantry units and move them into the vehicle's cargo positions
+private _infantryGroup = createGroup [_side, true];
+
+{
+    private _unit = _infantryGroup createUnit [_x, [0, 0, 0], [], 0, "CAN_COLLIDE"];
+    _unit moveInCargo _vehicle;
+} forEach _infantryTypes;
```
shukari: If function get to many cargo for the vehicle... the units are at [0,0,0]
```
{
    private _unit = _infantryGroup createUnit [_x, [0, 0, 0], [], 0, "CAN_COLLIDE"];
    _unit assignAsCargoIndex [heli, _forEachIndex];
    _unit moveInCargo _vehicle;

    if (_vehicle getCargoIndex _unit != _forEachIndex) then {deleteVehicle _unit};
} forEach _infantryTypes;
```
or
```
{
    private _unit = _infantryGroup createUnit [_x, [0, 0, 0], [], 0, "CAN_COLLIDE"];
    _unit assignAsCargo _vehicle;
    if !(_unit moveInAny _vehicle) then {deleteVehicle _unit};
} forEach _infantryTypes;
```

addons/modules/functions/fnc_moduleSpawnReinforcements.sqf
```diff
@@ -0,0 +1,136 @@
...
+    if (_isAir) then {
+        private _waypoint = _vehicleGroup addWaypoint [_positionLZ, WAYPOINT_RADIUS];
+        _waypoint setWaypointType "SAD";
+        _waypoint setWaypointBehaviour "AWARE";
+        _waypoint setWaypointCombatMode "RED";
+    };
+};
+
+// Add the vehicle and crew + passengers to curators
+[QEGVAR(common,addObjects), [[_vehicle]]] call CBA_fnc_localEvent;
```
shukari: return vehicle, crew and cargo would be nice if someone like me want to use it, but it is not public
`[_vehicle, _vehicleGroup, _infantryGroup]`

neilzar: The function is not supposed to be exposed to the public API, it only is called from the confirm button of the module. So no return value is needed.

addons/modules/functions/fnc_moduleSpawnReinforcements.sqf
neilzar: The module doesn't allow more units to be selected than spots are available in the vehicle, so this is a non-issue.

addons/modules/functions/fnc_moduleAmbientFlyby.sqf
neilzar: You can create a new PR if you want to change this, this has already been merged and the branch has been deleted.

PR #214, opened 2019-10-15
Improve object editing (vehicle in vehicle cargo, jet sounds setting)

mharis001:
- Improve `BIS_fnc_curatorObjectEdited` by:
    - Add support for loading/unloading vehicle in vehicle cargo
    - Improving handling of slingloaded objects
        - Properly handle situation where object is already slingloaded
        - Object will detach if moved (without having to hover over another object)
    - Use `setVehiclePosition` when moving objects to ground level
        - Performs check for collisions with other objects
    - Cleaning up code
- Add `zen_editor_parachuteSounds` setting
    - Controls whether jet flyby sounds are played when objects are given parachutes

### DIFF
--- a/addons/editor/stringtable.xml
+++ b/addons/editor/stringtable.xml
@@ -64,7 +64,7 @@
             <English>Parachute Sounds</English>
         </Key>
         <Key ID="STR_ZEN_Editor_ParachuteSounds_Description">
-            <English>Controls whether jet flyby sounds are played when object positions are changed to a height where they are attached to parachutes.</English>
+            <English>Controls whether jet flyby sounds are played when an object is paradropped via the Raise Entity keybind.</English>
         </Key>
         <Key ID="STR_ZEN_Editor_UnitRadioMessages_WaypointsOnly">
             <English>Waypoints Only</English>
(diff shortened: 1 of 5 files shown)


addons/editor/stringtable.xml
```diff
@@ -60,6 +60,12 @@
             <Polish>Kontroluje czy wiadomości radiowe jednostki są odtwarzane przy postawieniu i wyznaczeniu punktu trasy, jedynie przy wyznaczeniu punktu trasy lub czy są wyłączone.</Polish>
             <Russian>Определяет будут ли ИИ подразделения сообщать по радио о своих действиях или нет.</Russian>
         </Key>
+        <Key ID="STR_ZEN_Editor_ParachuteSounds">
+            <English>Parachute Sounds</English>
+        </Key>
+        <Key ID="STR_ZEN_Editor_ParachuteSounds_Description">
+            <English>Controls whether jet flyby sounds are played when object positions are changed to a height where they are attached to parachutes.</English>
```
CreepPork:
```suggestion
            <English>Controls whether jet flyby sounds are played when an object is para-dropped via ALT drag.</English>
```

mharis001: Maybe "via the Raise Entity keybind." since the key can be rebound (I don't know why anyone would).

CreepPork: I don’t know either but that could work.

PR #218, opened 2019-10-17
Disable Apply To All button when no other vehicles are selected

mharis001:
- title

### DIFF
--- a/addons/garage/stringtable.xml
+++ b/addons/garage/stringtable.xml
@@ -18,5 +18,8 @@
         <Key ID="STR_ZEN_Garage_ApplyToAll_Tooltip">
             <English>Apply the current vehicle's customization to all selected vehicles of this type.</English>
         </Key>
+        <Key ID="STR_ZEN_Garage_CannotApplyToAll">
+            <English>No other vehicles of the same type selected.</English>
+        </Key>
     </Package>
 </Project>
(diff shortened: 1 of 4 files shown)


addons/garage/stringtable.xml
```diff
@@ -18,5 +18,8 @@
         <Key ID="STR_ZEN_Garage_ApplyToAll_Tooltip">
             <English>Apply the current vehicle's customization to all selected vehicles of this type.</English>
         </Key>
+        <Key ID="STR_ZEN_Garage_CannotApplyToAll">
+            <English>No other vehicles of the same type selected.</English>
```
CreepPork: Have been selected.

neilzar: In the context where the string is used the current form is correct.

CreepPork: English is not my native language, but to me, it just sounds a tad off.

neilzar: I'm not saying it as a case of personal preference, I'm saying it as a case of correct grammar. Past perfect is not the correct form to use in this context, there is no sequence of events that would require it. I had a bilingual education, so I should know this.

jonpas: While I agree, there is something about messages being short and clear, and la language does get in the way sometimes. :)

PR #219, opened 2019-10-18
Add Deep Copy/Paste keybinds

mharis001:
- Add Deep Copy and Deep Paste keybinds
- Add `getPylonTurret` common function
- Update `getPosFromScreen` common function
    - Add `_checkIntersections` parameter
    - Allow multiple intersections from the same object
    - Increase `MAX_RESULTS` to 5


addons/editor/functions/fnc_deepPaste.sqf
```diff
@@ -0,0 +1,221 @@
...
+
+    {
+        _x params ["_unitData", "_role", "_cargoIndex", "_turretPath"];
+
+        private _unit = _unitData call _fnc_createUnit;
+
+        switch (_role) do {
+            case "driver": {
+                if (getText (configFile >> "CfgVehicles" >> typeOf _unit >> "simulation") == "UAVPilot") then {
+                    _unit moveInAny _vehicle; // moveInDriver does not appear to work for virtual UAV crew, moveInAny does
```
CreepPork: Will it guarantee that it will always be the first unit? Because the rover UAV has slots for the passengers as well and if the unit is moveInAny’d then it may place the UAV unit in the passenger seat, no?

mharis001: `moveInAny` prioritizes the driver role and the other crew are placed using role specific commands.

addons/editor/functions/fnc_deepCopy.sqf
```diff
@@ -0,0 +1,154 @@
...
+
+// Get the world position where the cursor is pointing
+// Object positions are copied relative to the cursor
+private _centerPos = if (visibleMap) then {
+    private _ctrlMap = findDisplay IDD_RSCDISPLAYCURATOR displayCtrl IDC_RSCDISPLAYCURATOR_MAINMAP;
+    private _pos2D = _ctrlMap ctrlMapScreenToWorld getMousePosition;
+    _pos2D + [0]
+} else {
+    screenToWorld getMousePosition
+};
```
CreepPork: Should be extracted into a new function.

addons/editor/functions/fnc_deepCopy.sqf
```diff
@@ -0,0 +1,154 @@
...
+    private _index = _indexedGroups find _group;
+
+    if (_index == -1) then {
+        _index = _indexedGroups pushBack _group;
+
+        _groups pushBack [side _group, formation _group, behaviour leader _group, combatMode _group, speedMode _group];
+    };
+
+    _index
+};
```
CreepPork: Maybe extract this whole feature to a new component and make each of these functions separate? Would make this main code much cleaner.

addons/editor/initKeybinds.sqf
```diff
@@ -21,6 +21,24 @@
...
 
-        true // handled, prevents vanilla eject from activating
+        true // handled, prevents vanilla eject
     };
 }, {}, [DIK_G, [false, true, false]]] call CBA_fnc_addKeybind; // Default: CTRL + G
+
+[ELSTRING(common,Category), QGVAR(deepCopy), [LSTRING(DeepCopy), LSTRING(DeepCopy_Description)], {
+    if (!isNull curatorCamera && {!GETMVAR(RscDisplayCurator_search,false)}) then {
+        GVAR(clipboard) = [SELECTED_OBJECTS] call FUNC(deepCopy);
+        playSound ["RscDisplayCurator_error01", true];
```
CreepPork: Why is it playing the error sound? It could play the `FD_Finish_F` for better UX effect.

addons/editor/initKeybinds.sqf
```diff
@@ -21,6 +21,24 @@
...
+
+        true // handled, prevents vanilla copy
+    };
+}, {}, [DIK_C, [true, true, false]]] call CBA_fnc_addKeybind; // Default: CTRL + SHIFT + C
+
+[ELSTRING(common,Category), QGVAR(deepPaste), [LSTRING(DeepPaste), LSTRING(DeepPaste_Description)], {
+    if (!isNull curatorCamera && {!GETMVAR(RscDisplayCurator_search,false)}) then {
+        [QGVAR(deepPaste), GVAR(clipboard)] call CBA_fnc_serverEvent;
+        playSound ["RscDisplayCurator_error01", true];
```
CreepPork:
```suggestion
        playSound ["FD_Finish_F", true];
```

addons/editor/script_component.hpp
```diff
@@ -2,9 +2,9 @@
 #define COMPONENT_BEAUTIFIED Editor
 #include "\x\zen\addons\main\script_mod.hpp"
 
-// #define DEBUG_MODE_FULL
-// #define DISABLE_COMPILE_CACHE
-// #define ENABLE_PERFORMANCE_COUNTERS
+#define DEBUG_MODE_FULL
+#define DISABLE_COMPILE_CACHE
+#define ENABLE_PERFORMANCE_COUNTERS
```
CreepPork: When out of draft state, should be commented out.

addons/editor/functions/fnc_deepCopy.sqf
mharis001: Was going to use `getPosFromScreen` after making the necessary changes to it.

addons/editor/initKeybinds.sqf
mharis001: Despite the name, this is the ping sound played when a Zeus message is shown.

addons/editor/functions/fnc_deepCopy.sqf
mharis001: While I do agree that it would nicer for these individual serializing functions to be in a separate file, I think the use of the local variables such as `_centerPos` and `_indexedGroups` might become cumbersome as these would then need to be passed to the respective functions rather than relying on them being defined.

addons/common/XEH_postInit.sqf
```diff
@@ -233,6 +233,16 @@
...
 [QGVAR(setDate), {setDate _this}] call CBA_fnc_addEventHandler;
 
+[QGVAR(setUnitIdentity), {
+    params ["_unit", "_name", "_face", "_speaker", "_pitch", "_nameSound"];
+
+    _unit setName _name;
+    _unit setFace _face;
+    _unit setSpeaker _speaker;
+    _unit setPitch _pitch;
+    _unit setNameSound _nameSound;
```
CreepPork: From Achilles and personal experiences, these should be a keybind option or a CBA option to copy with such depth. As some players may want to copy the load-outs and not make the entire group of clones.

mharis001: With this being a "deep copy", I would expect "everything" to be copied. I don't think a setting is necessary, however, this can always be re-assessed later based on feedback.

CreepPork: It depends on the situation. I have, sometimes, experienced that I need only their gear not all of them to share voices, faces, etc. but, on the other hand, sometimes you just need a full clone (e.g. the specific officer was killed accidentally and you need to dupe him).

CreepPork: I guess we could merge it for now and in a later update, devise a solution for this issue.

PR #226, opened 2019-10-20
Improve garage tabs handling

mharis001:
- Garage reopens with previously selected tab
- Selected tab can be deselected by clicking the button again or left clicking
- Allow using `TAB` key to cycle between tabs

### DIFF
--- a/addons/garage/XEH_preInit.sqf
+++ b/addons/garage/XEH_preInit.sqf
@@ -6,6 +6,7 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+GVAR(currentTab) = -1;
 GVAR(helperPos) = [0, 0, -1];
 GVAR(camDistance) = 100;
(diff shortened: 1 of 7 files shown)


addons/garage/XEH_preInit.sqf
```diff
@@ -6,6 +6,7 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+GVAR(currentTab) = -1;
```
CreepPork: By default, why shouldn’t it start at 0?

mharis001: I think I chose this since BI garage initially opens with no tab selected.

mharis001:
Can't reproduce, working fine on my end after 20+ attempts.

neilzar:
Happens every time for me, running only CBA and ZEN. Build using `build.py`.

mharis001:
Determined to result from `unsafeCVL` changes in development branch. Unrelated to changes in this PR.

PR #240, opened 2019-11-04
Fix group trees preventing editable icon selection

mharis001:
- Fix selecting a non-placement path in group trees preventing the selection of editable icons using left click
- This makes the group trees match the behaviour of other trees by allowing selection to still occur
- Without this, path has to be deselected by either right clicking or selecting an entity using a selection box

### DIFF
--- a/addons/editor/functions/fnc_initDisplayCurator.sqf
+++ b/addons/editor/functions/fnc_initDisplayCurator.sqf
@@ -18,10 +18,6 @@
 
 params ["_display"];
 
-// Reset editable icons visibility tracking variable
-// Prevents unwanted behaviour if display is closed while icons are hidden
-GVAR(iconsVisible) = true;
-
 if (GVAR(removeWatermark)) then {
     private _ctrlWatermark = _display displayCtrl IDC_RSCDISPLAYCURATOR_WATERMARK;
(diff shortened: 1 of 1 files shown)


addons/editor/functions/fnc_initDisplayCurator.sqf
```diff
@@ -73,6 +69,25 @@ _display displayAddEventHandler ["KeyDown", {call FUNC(handleKeyDown)}];
     }];
 } forEach IDCS_SIDE_BUTTONS;
 
+// Fix selecting a path in group trees preventing selecting an editable icon using left click
+// This matches the behaviour of other trees by deselecting any non-placement path when selecting an editable icon
+{
+    private _ctrl = _display displayCtrl _x;
+    _ctrl ctrlAddEventHandler ["MouseButtonDown", {
+        if (RscDisplayCurator_sections select 0 == 1) then {
```
CreepPork: Wasn’t there a function or global var already for this as a shortcut?

mharis001: If you mean for: `RscDisplayCurator_sections select 0` - no, not that I know of. This is the only way to get the current mode.

PR #241, opened 2019-11-05
Add Toggle Lamps module

mharis001:
- title


addons/common/XEH_PREP.hpp
```diff
@@ -15,6 +15,7 @@ PREP(getInventory);
 PREP(getPhoneticName);
 PREP(getPlayers);
 PREP(getPosFromScreen);
+PREP(getReflectorSelections);
```
CreepPork: In my opinion, getLightingHitpoints or something similar is quicker to understand than this.

addons/common/functions/fnc_setLampState.sqf
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Arguments:
+ * 0: Lamp <OBJECT>
+ * 1: State <BOOL>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [_lamp, false] call zen_common_fnc_setLampState
```
CreepPork: I think there is already a BI function for this. Any reason for adding our own?

jonpas: Do a comparison of them. General rule of thumb for ACEifying anything is to not rely on BI functions, they may be written in a worse way, without the given framework for debugging, or have additional checks that slow it down and are not necessary.

CreepPork: BI's
```sqf
params [
	["_lamp",objnull,[objnull]],
	["_state",true,[true]]
];
_lamp switchlight (["off","on"] select _state);
```

It uses `switchLight` instead of `setHit` but it seems that in the latest A3 updates `switchLight` has been extended to allow it to be used on street lamps too. Don't see the point of using ZEN's implementation.

mharis001: Own function is better here, it will handle destroyed lights.

CreepPork: I'm not trying to advocate for BI's code, but handling destroyed lights might not be what the user intends. A player might shoot out lights and then using this module would repair them all back to normal. A choice would be good.

PR #243, opened 2019-11-07
Add Editable Objects context menu action

### LINKED ISSUE #178: Context menu option: Update Editable Objects
When trying to manage objects in zeus, updating editable objects can become a slight chore to find and place the module every time.

**Solution you'd like:**
It would be nice to have a default context menu action for the "Update Editable Objects" function to accompany the "Teleport Zeus" action.

mharis001:
- title, Close #178


docs/user_guide/context_actions.md
```diff
@@ -27,6 +27,10 @@ Sub-actions allow for copying and pasting the hovered units' loadout onto anothe
 Opens the Zeus garage on the hovered vehicle.
 Sub-actions allow for copying and pasting the hovered vehicle's customization onto another.
 
+# Editable Objects
+
+Allows Zeus to add or remove editable objects within the selected radius.
```
neilzar:
```suggestion
Allows the Zeus to add or remove editable objects within the selected radius.
```

mharis001: I think no "the" is correct as there can be multiple, it is like this in other descriptions as well.

mharis001:
The position is taken from the camera to the  top-left corner of the menu. Are you moving it after opening?

docs/user_guide/context_actions.md
neilzar: I always say Zeuses for multiple.

neilzar:
No, camera stays in the same position.

docs/user_guide/context_actions.md
jonpas: To me "Allow Zeus" reads smoother than "Allow the Zeus". But that might just be because I got used to it.

mharis001:
The actions work, I think it can be difficult to judge the distance and changing to 2D mode should help with that.

addons/context_actions/CfgContext.hpp
```diff
@@ -290,6 +290,71 @@ class EGVAR(context_menu,actions) {
...
+    class EditableObjects {
+        displayName = CSTRING(EditableObjects);
+        icon = QPATHTOEF(modules,ui\edit_obj_ca.paa);
+        priority = -98;
+        class Add {
+            displayName = ECSTRING(common,Add);
+            icon = QPATHTOF(ui\add_ca.paa);
+            priority = 2;
+            class 10m {
+                displayName = "10 m";
```
CreepPork: The `m` should be localized for other languages.

neilzar: m is the SI symbol for meter, it is universal.

mharis001: There should be a space between the numerical value and the unit symbol.

CreepPork: The `m` is an SI symbol but that doesn't make it fully universal. For example, Crylic would use м, not m. Granted, in this case, the difference is minor but something like kg it would be far more important (кг). It's good to start localizing them here and keep it simplified in the long run.

addons/context_actions/CfgContext.hpp
```diff
@@ -299,27 +299,27 @@ class EGVAR(context_menu,actions) {
             icon = QPATHTOF(ui\add_ca.paa);
             priority = 2;
             class 10m {
-                displayName = "10 m";
+                displayName = CSTRING(10m);
```
CreepPork: Couldn't have this just used the `FORMAT` macro and only localized the meter (moving it to common stringtable or something)?

mharis001: No, you have to localize the entire string, cannot use a combination. That macro is essentially the `format` command with the suffix `_#` as the number of arguments.

jonpas: So `format [CSTRING(m), 10]` and string `"%1 m` only works in SQF?

mharis001: AFAIK you can only do that in SQF.

PR #246, opened 2019-11-07
Add meta.cpp file, steam workshop support

mharis001:
- title, [reference](https://community.bistudio.com/wiki/How_to_enable_Steam_Workshop_support_in_Server_Browser)


CreepPork:
Isn’t meta.cpp already generated by the Publisher?

mharis001:
It is, I think is needed if someone does not download from the workshop. I am not sure though.

dedmen:
> if someone does not download from the workshop.

and then loads that mod on the server, yes. Thats needed then, otherwise the server won't detect the workshop item ID.

PR #248, opened 2019-11-08
Add right click to clear to engine search bar

mharis001:
- title, same as with custom search bar


dedmen:
a bit unrelated, but I saw that "backspace" in search field doesn't work properly. But I guess you cannot solve that can you?

mharis001:
What do you mean by doesn't work properly?

PR #251, opened 2019-11-12
Use common events instead of display XEH

mharis001:
- Use common `ZEN_displayCuratorLoad` event instead of display XEH
    - Unify code, some components used this event others added a display XEH
- Rename `curatorDisplayLoad` and `curatorDisplayUnload` common functions to match event names


bux:
imho, past tense for the event names would make them sound better.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members?redirectedfrom=MSDN#names-of-events

dedmen:
> past tense for the event names would make them sound better.

Its not just about sound, its also about meaning.
OnCuratorLoad should fire before load, OnCuratorLoaded should fire after load is done.

bux:
> Its not just about sound, its also about meaning.

that's why I've added the ms documentation :)

> OnCuratorLoad 

`OnCuratorLoading`

mharis001:
Unfortunately, these events are already existing public ones. Other events such as `remoteControlStarted` or `modeChanged` do follow that convention.

CreepPork:
We could deprecate the old names for the events and then in 1.8.0 remove the old ones. It's better to keep it all in check, even though it will be backwards incompatible.

mharis001:
I don't think breaking backwards compatibility only to change naming is worth it. Either way, these event names are what Arma uses - `onLoad`, `onUnload` - just prefixed for ZEN.

CreepPork:
Later, we'll end up with more of these things that are out of sync with the guidelines and it will become harder to change.

We aren't really breaking anything much if we offer a version with the new event name and the old one deprecated. Then one version later after this is merged, we can remove the old name. Keeps things simple and clean, without immediately breaking everything (it gives a clean upgrade path).

dedmen:
> We aren't really breaking anything much if we offer a version with the new event name and the old one deprecated.

I kinda did something similar with TFAR when I moved over from TFAR proprietary eventhandlers to CBA.

PR #254, opened 2019-11-17
Add Custom Fire module

### LINKED ISSUE #265: Custom Fire module (similar to Achilles)

mharis001:
- Add module to create a custom fire effect
- Similar to 3DEN equivalent but can be edited after creation and has sound effect
- Close #265


addons/modules/functions/fnc_moduleEffectFire.sqf
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * Public: No
+ */
+
+params ["_display"];
+
+private _logic = GETMVAR(BIS_fnc_initCuratorAttributes_target,objNull);
+_display closeDisplay IDC_CANCEL; // Close helper display
+
+// Need to delay dialog creation by one frame to avoid weird input blocking bug
```
CreepPork: What about other dialogs? Does it only impact this one?

docs/user_guide/modules_list.md
```diff
@@ -100,6 +100,10 @@ Creates a teleport location at the attached object or, if not attached, at a new
 Players can use the "Teleport" action on any teleporter objects to teleport between them.
 When used on a vehicle, players will first be teleported into an empty seat if possible.
 
+## Custom Fire
+
+Creates a custom fire effect based on the selected properties.
```
CreepPork: A bit too vague for the number of options present.

addons/modules/functions/fnc_moduleEffectFire.sqf
mharis001: This is from the light source module which uses the same method. I think it results from closing and creating a new dialog in the same frame.

PR #272, opened 2019-12-21
Fix paradrop waypoint for Spawn Reinforcements module

### LINKED ISSUE #262: Spawn Reinforcements: Slow down aircraft when approaching an LZ
Not sure if this is a real bug report. I had some troubles when using the paradrop function with an MH-6. When approaching the LZ the velocity of the MH-6 was too high and it always overshot the LZ having to do circles again and again. When manually setting it to limited speed it was able to successfully paradrop the reinforcements. The same is happening when letting a heli fastrope its units or land. Or when a plane (blackfish infantry transport) is trying to paradrop units.

**Solution you'd like:**
Add a waypoint with the "limited" speed setting a few 100 meters before the actual LZ.

Having an aircraft always fly at limited speed. But story telling wise that might not always be preferred.

mharis001:
- title, SAD waypoint makes aircraft slow down when approaching previous waypoint, adding the SAD waypoint after the insertion one completes seems like the most reliable way to avoid this issue
- Close #262

### DIFF
--- a/addons/ai/functions/fnc_waypointParadrop.sqf
+++ b/addons/ai/functions/fnc_waypointParadrop.sqf
@@ -16,7 +16,7 @@
  * Public: No
  */
 
-#define MOVE_DELAY 3
+#define MOVE_DELAY 2
 #define OFFSET_DISTANCE 2000
 #define TIME_PER_UNIT 1.2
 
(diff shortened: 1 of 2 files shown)


addons/ai/functions/fnc_waypointParadrop.sqf
```diff
@@ -55,12 +55,15 @@ waitUntil {
 
     // Check if the aircraft is close enough to the waypoint to begin paradropping
     // Distance is based on the current speed of aircraft and the number of passengers
-    private _passengers = {assignedVehicleRole _x select 0 == "cargo"} count crew _vehicle;
+    _passengers = {assignedVehicleRole _x select 0 == "cargo"} count crew _vehicle;
```
CreepPork:
```suggestion
    _passengers = {(assignedVehicleRole _x) select 0 == "cargo"} count crew _vehicle;
```
Personal preference to increase readability of this statement.

mharis001: I prefer not to have parentheses around unary commands. This line reads naturally from left-to-right either way IMO.

bux: agree with @CreepPork

PR #275, opened 2019-12-27
Add Export Mission SQF module

### LINKED ISSUE #220: Feature Request: Copy Mission SQF
People in my community like to make a mission beforehand, copy the mission into an SQF using Achilles' `Copy to Mission SQF`, and then run the SQF at the start of the session so all objects and markers are pre-placed.

**Solution you'd like:**
A module to copy the mission objects, markers, etc. to an SQF that can be run using the `Execute Code` module

mharis001:
- Add "Export Mission SQF" module, Close #220 
- Add `exportMissionSQF` common function
- Add `exportText` common function - opens an export dialog with copy button if possible
- Add `hasDefaultInventory` common function


addons/common/functions/fnc_exportMissionSQF.sqf
```diff
@@ -0,0 +1,381 @@
...
+    ["_radius", -1, [0]],
+    ["_includeWaypoints", true, [true]],
+    ["_includeMarkers", false, [true]],
+    ["_addToCurators", false, [true]]
+];
+
+private _processedObjects = [];
+private _processedGroups  = [];
+
+private _outputGroups1   = [];
```
CreepPork: Ew, bad names.

addons/common/functions/fnc_exportMissionSQF.sqf
```diff
@@ -0,0 +1,381 @@
...
+private _outputObjects   = [];
+private _outputGroups2   = [];
```
CreepPork: Same here.

addons/common/functions/fnc_exportMissionSQF.sqf
```diff
@@ -0,0 +1,381 @@
...
+            _outputMarkers pushBack ["%1 setMarkerBrush %2;", _varName, str markerBrush _x];
+            _outputMarkers pushBack ["%1 setMarkerColor %2;", _varName, str markerColor _x];
+            _outputMarkers pushBack ["%1 setMarkerAlpha %2;", _varName, markerAlpha _x];
+            _outputMarkers pushBack "";
+        };
+    } forEach allMapMarkers;
+};
+
+if (_addToCurators) then {
+    // todo
```
CreepPork: ToDo

addons/common/functions/fnc_exportMissionSQF.sqf
mharis001: Not sure what would be better, open to suggestions.

addons/common/functions/fnc_exportMissionSQF.sqf
```diff
@@ -0,0 +1,381 @@
...
+            _outputMarkers pushBack ["%1 setMarkerShape %2;", _varName, str markerShape _x];
+            _outputMarkers pushBack ["%1 setMarkerSize %2;", _varName, markerSize _x];
+            _outputMarkers pushBack ["%1 setMarkerText %2;", _varName, str markerText _x];
+            _outputMarkers pushBack ["%1 setMarkerBrush %2;", _varName, str markerBrush _x];
+            _outputMarkers pushBack ["%1 setMarkerColor %2;", _varName, str markerColor _x];
+            _outputMarkers pushBack ["%1 setMarkerAlpha %2;", _varName, markerAlpha _x];
+            _outputMarkers pushBack "";
+        };
+    } forEach allMapMarkers;
+};
```
neilzar: Should also include the area markers if the PBO is not removed.

mharis001: This will include area markers, only polylines are excluded.

addons/common/functions/fnc_exportMissionSQF.sqf
neilzar: Is this even needed? There is no option to change this in the module, and there is the "Auto Add Object" option that also does this.

addons/common/functions/fnc_exportMissionSQF.sqf
```diff
@@ -0,0 +1,381 @@
...
+            "CAN_COLLIDE"
+        };
+
+        _outputObjects pushBack ["%1 = createVehicle [%2, [0, 0, 0], [], 0, %3];", _varName, str typeOf _vehicle, str _placement];
+        _outputObjects pushBack ["%1 setVectorDirAndUp %2;", _varName, [vectorDir _vehicle, vectorUp _vehicle]];
+        _outputObjects pushBack ["%1 setPosASL %2;", _varName, getPosASL _vehicle];
+        _outputObjects pushBack ["%1 setDamage %2;", _varName, damage _vehicle];
+        _outputObjects pushBack ["%1 setFuel %2;", _varName, fuel _vehicle];
+
+        _outputObjects pushBack ["{%1 setHitIndex [_forEachIndex, _x, false]} forEach %2;", _varName, getAllHitPointsDamage _vehicle select 2];
```
neilzar: Zero Divisor error on the `select 2`. Happens every time, even on an empty mission. Haven't been able to figure out what exactly causes it.

allMissionObjects "All": 
```
[L Modules:1, Agent 0x7a22c080, Agent 0x348b8180, Agent 0x7a2ac100, L Zeus:1, henk, 2899531: cl_feathers2.p3d, 2899517: cl_grassmulti.p3d, 2899534: cl_feathers2.p3d, 2899526: cl_feathers2.p3d, Agent 0x79564100, Agent 0x69aee140]
```

addons/common/functions/fnc_exportMissionSQF.sqf
```diff
@@ -0,0 +1,381 @@
...
+            _object call _fnc_processVehicle
+        };
+    } else {
+        if (_object isKindOf "Thing" || {_object isKindOf "Static"}) then {
+            _object call _fnc_processStatic
+        };
+    };
+};
+
+private _objects = allMissionObjects "All";
```
neilzar:
```suggestion
private _objects = curatorEditableObjects getAssignedCuratorLogic player - [objNull];
```
Gives a cleaner output without objects that can't be edited. Might solve the select error on line 221.

Empty mission Array size on Livonia with only one Zeus module.
Previous: 12
Suggested change: 1

With 8 Rhinos placed down:
Previous: 13 😕 
Suggested: 9

With 13 Rhinos placed down:
Previous: 19
Suggested: 14

addons/common/functions/fnc_exportMissionSQF.sqf
mharis001: Idea was to have the outputted SQF add all created objects to curators independently of the `autoAddObjects` setting.

addons/common/functions/fnc_exportMissionSQF.sqf
mharis001: With this, objects that are not added to the local curator would not be exported.

addons/common/functions/fnc_exportMissionSQF.sqf
mharis001: Results from `getAllHitPointsDamage` returning `[]` for some objects.

addons/common/functions/fnc_exportMissionSQF.sqf
neilzar: I don't see why you would want to export those, I have never encountered a situation where I needed to export objects that were not added to my curator. And you can add objects to the curator using the add to curator action/module if needed.

addons/common/functions/fnc_exportMissionSQF.sqf
mharis001: Removed for now, can be added later if needed.

addons/common/functions/fnc_exportMissionSQF.sqf
mharis001: To me it does not make sense limit the function to editable objects. By "without objects that can't be edited", do you mean objects that are not _currently_ editable?

neilzar: Where did I say that?

neilzar: In my community, it is even beneficial to only copy the objects that are added to the curator. We use templates with tents where the player prep for the op, and we don't want to copy those because they will already be on the template when pasting the code. I think that my suggested version is better because that way you don't have to add editor placed objects and then delete them to not copy them; you just don't add them.

addons/common/functions/fnc_exportMissionSQF.sqf
```diff
@@ -0,0 +1,421 @@
...
+#define FORMAT_OBJ_VAR(index) format ["_object%1", index]
+#define FORMAT_GRP_VAR(index) format ["_group%1", index]
+
+#define NEWLINE toString [ASCII_NEWLINE]
+
+params [
+    ["_position", [0, 0, 0], [[]], [2, 3]],
+    ["_radius", -1, [0]],
+    ["_includeWaypoints", true, [true]],
+    ["_includeMarkers", false, [true]],
```
neilzar:
```suggestion
    ["_includeMarkers", false, [true]]
```

addons/common/functions/fnc_exportMissionSQF.sqf
neilzar: To give another reason for the switch. On an empty mission, where `allMissionObjects "All"` returns only three objects, I get the following timings:
`allMissionObjects "All"`: 11.47ms
The new suggestion: 0.0011ms

mharis001: I can see the use in this. I added an option for `exportMissionSQF` to export curator editable objects only - think that is the best solution.

neilzar: That is a good solution.

mharis001:
Can add another filter for `ThingEffect` or maybe filtering all objects that are not `scope == 2` might be better?

neilzar:
That is probably the easiest solution.

mharis001:
Filtering by scope filters UAV crew, changed to just filter `ThingEffect`.

addons/common/functions/fnc_exportMissionSQF.sqf
```diff
@@ -0,0 +1,428 @@
...
+// Keep track of all processed objects and groups, their index in their corresponding array
+// is used to determine the variable name used in the exported SQF
+private _processedObjects = [];
+private _processedGroups = [];
+
+// Separate the exported SQF into different sections, this is used to ensure the correct
+// ordering of the output (for example, applying group properties after all units are created)
+private _outputGroups1 = [];
+private _outputObjects = [];
+private _outputGroups2 = [];
```
CreepPork: Maybe some comments to describe why are they named so similarly?

addons/common/functions/fnc_hasDefaultInventory.sqf
```diff
@@ -0,0 +1,88 @@
...
+    private _current = getUnitLoadout _object;
+    private _default = getUnitLoadout typeOf _object;
+
+    // Always match facewear because it is influenced by identity, not config inventory
+    _default set [7, _current select 7];
+
+    // Sort uniform, vest, and backpack item arrays, their contents can be indentical
+    // but the arrays may be different depending on the order they were added to the container
+    {
+        for "_i" from 3 to 5 do {
```
CreepPork: Magic numbers, at least a comment.

addons/common/functions/fnc_exportMissionSQF.sqf
mharis001: L43 explains why there are 2 group arrays.

PR #281, opened 2019-12-30
Fix drawIcon3D icons being hidden after using garage

mharis001:
- title


CreepPork:
Are both fixes (#255 and this) required?

mharis001:
Yes, I originally thought it was only Virtual Arsenal, but, I guess its for any created camera.

PR #289, opened 2020-01-13
Fix displaying Execute Code module's JIP ID

mharis001:
- title, execute `setName` on all clients
- Use CBA events instead of `remoteExec`

### DIFF
--- a/addons/modules/functions/fnc_gui_executeCode.sqf
+++ b/addons/modules/functions/fnc_gui_executeCode.sqf
@@ -101,8 +101,12 @@ private _fnc_onConfirm = {
             [QEGVAR(common,execute), [_code, _args]] call CBA_fnc_globalEvent;
         };
         case 3: {
-            private _jipID = [_args, _code] remoteExec ["call", 0, _logic];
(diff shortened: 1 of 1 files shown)


addons/modules/functions/fnc_gui_executeCode.sqf
```diff
@@ -101,8 +101,12 @@ private _fnc_onConfirm = {
             [QEGVAR(common,execute), [_code, _args]] call CBA_fnc_globalEvent;
         };
         case 3: {
-            private _jipID = [_args, _code] remoteExec ["call", 0, _logic];
-            _logic setName format ["JIP ID: %1", _jipID];
+            private _jipID = [QEGVAR(common,execute), [_code, _args]] call CBA_fnc_globalEventJIP;
+            [_jipID, _logic] call CBA_fnc_removeGlobalEventJIP;
```
CreepPork: I think I'm missing something but why create a JIP event and then right after you remove it?

mharis001: It's removed after the logic is deleted.

PR #316, opened 2020-03-07
Add release drafter CI

neilzar:
- _Title_

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,34 @@
+branches:
+  - master
+
+name-template: 'Version $NEXT_PATCH_VERSION'
+tag-template: 'v$NEXT_PATCH_VERSION'
+
+categories:
+  - title: '**ADDED:**'
+    labels: 'feature'
+  - title: '**FIXED:**'
+    labels: 'bug fix'
+  - title: '**IMPROVED**'
+    labels:
+      - 'enhancement'
+      - 'optimization'
+  - title: '**CHANGED**'
+    labels: 'cleanup'
+  - title: '**TRANSLATIONS**'
+    labels: 'translation'
+
+exclude-labels: 'ignore changelog'
+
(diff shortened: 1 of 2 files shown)


CreepPork:
Could you add an example generated changelog from the drafter (e.g. v1.7.0)?

I'm worried that it might be too unclean to be of any use where you'd have to go and still edit every single line where it might be quicker to do it by hand altogether.

jonpas:
Example: https://github.com/acemod/ACEX/releases/tag/v3.5.1 (the only edited thing is required CBA version)

Release Drafter still has some bugs, it will duplicate entries for PRs with multiple tags. And it also works based on date alone, so hotfixes will erase it (easy to overcome if you just copy it somewhere else). It will also always use only last draft, no matter who made it.

And finally, it was very useful for latest ACE3 and ACEX releases, you just need a standard for PR titles to keep it clean.

neilzar:
> Release Drafter still has some bugs, it will duplicate entries for PRs with multiple tags. 
This was apparently already fixed in the latest release, 26 days ago.

I can't find an issue on the release-drafter repo about the erasing issue.

jonpas:
Because it's not an issue, it's how it works. It will always draft from latest release to current.

.github/release-drafter.yml
```diff
@@ -1,34 +1,36 @@
...
 
 change-template: '- $TITLE ($NUMBER)'
 template: |
-  _Requires CBA version X.Y.Z or later and Arma 3 version X.Y or later._
+  _Requires CBA Version X.Y.Z or later and Arma 3 Version X.Y or later._
+
+  DESCRIPTION HERE
 
-  ## Changelog Summary
+  ## Change Log Summary
```
neilzar, line 29: Changelog is actually the spelling used on wiki and in many other places I encounter. I don't see Change Log being used as much.

mharis001: It has been used in all releases so far.

PR #322, opened 2020-03-12
Add button to randomize custom compositions

### LINKED ISSUE #321: Stop deep copy/paste & custom compositions from altering face.
Deep copying a unit or making a composition of a unit or a group of units also copies the face, so you end up with an army of clones if you use it a lot.

**Solution you'd like:**
Stop deep copies/pastes and custom compositions from also setting the face. Maybe a toggle in the cba settings so people can choose for themselves. 
Maybe also do this for facewear? Some units have random facewear.

mharis001:
- Add randomize button to custom compositions panel
- Can be clicked to toggle randomization on/off for unit identities and vehicle customization when placing custom compositions
- Close #321


mharis001:
Merge #323 before merging this.

PR #342, opened 2020-03-27
Feature: area marker visibility by player side

Fusselwurm:
Add a new feature to the area marker configuration.

By default, new area markers will be visible to all players as before: 

But markers can be made invisible to some or all sides, like here - marker is only visible to blufor and civilian players:


Fusselwurm:
I have a weird issue. The way I implemented this is by exchanging `setMarkerAlpha` (run by Zeus) with `setMarkerAlphaLocal` (run by all clients). 

Works well, except for one thing: 
* given a marker that is visible to Zeus, and invisible to another client
* when Zeus doubleclicks the marker (to open the configuration)
* on the other client, the marker's alpha value mysteriously changes from 0 to either 1 (or to whatever the Zeus sees, not sure yet) ⚡️

From what I can see logging and looking into the profiler, neither the `applyProperties` nor the `updateAlpha` method are not called, so... looks like our code is not to blame.

My suspicion: engine somehow re-broadcasts global marker when I touch it.

Tentative conclusion: having a global marker and then running `*Local` commands on them invites trouble.

Will refactor to exclusively use local marker, see if that works better 😬

mharis001:
The `updateAlpha` event is currently only added on the server.

Fusselwurm:
> The `updateAlpha` event is currently only added on the server.

ah right, had forgotten to push the fix to that. 😓 the locality problem I described exists regardless.

I just refactored the whole thing to use local markers only, just tested with a  buddy and it seems to work. 👍 

things I still have to test : 
- [ ] with more than one zeus
- [ ] dragging markers

mharis001:
It happens because `onMouseButtonUp` is called when the icon is double clicked, which calls `setMarkerPos` and it seems that any global marker commands will re-synchronize all marker properties.

Fusselwurm:
aaah. right. ok. - so are you okay with the "let's be safe and do local markers only" solution, or should I keep global markers and try and change the `setMarkerPos` to `setMarkerPosLocal` ? @mharis001

mharis001:
I think the simpler solution is probably better - if the global `setMarkerPos` is the only issue then that can be changed. However, if there are other problems, local markers would make sense then.

addons/area_markers/functions/fnc_updateAlpha.sqf
```diff
@@ -0,0 +1,34 @@
+#include "script_component.hpp"
+/*
+ * Author: Fusselwurm
+ * Set the alpha of a marker depending on the player's side
```
CreepPork, line 4:
```suggestion
 * Set the alpha of a marker depending on the player's side.
```

addons/area_markers/functions/fnc_updateAlpha.sqf
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Arguments:
+ * 0: Marker <STRING>
+ * 1: Sides that may see the marker <ARRAY>
+ * 2: Alpha value to use for players of passed `sides`
```
CreepPork, line 9: Missing type

addons/area_markers/functions/fnc_updateAlpha.sqf
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * ["marker_0", [west,civilian], 0.7] call zen_area_markers_fnc_updateAlpha
```
CreepPork, line 15:
```suggestion
 * ["marker_0", [west, civilian], 0.7] call zen_area_markers_fnc_updateAlpha
```

addons/area_markers/functions/fnc_updateAlpha.sqf
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+params ["_marker", "_sides", "_alpha"];
+
+if (isServer) exitWith {
+    [GVAR(markerVisibilities), _marker, _sides] call CBA_fnc_hashSet;
+    publicVariable QGVAR(markerVisibilities);
+};
+
+if (!hasInterface) exitWith {}; // ignore HCs
```
CreepPork, line 27:
```suggestion
if (ZEN_isHC) exitWith {};
```

addons/area_markers/functions/fnc_updateMarkerPos.sqf
```diff
@@ -0,0 +1,21 @@
+#include "script_component.hpp"
+/*
+ * Author: Fusselwurm
+ * Set the marker position locally
```
CreepPork, line 4:
```suggestion
 * Set the marker position locally.
```

addons/area_markers/functions/fnc_updateMarkerPos.sqf
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Arguments:
+ * 0: Marker <STRING>
+ * 1: Marker position
```
CreepPork, line 8: Missing type

addons/area_markers/functions/fnc_updateMarkerPos.sqf
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * ["marker_0", [3265.59,853.12]] call zen_area_markers_fnc_updateMarkerPos
```
CreepPork, line 14:
```suggestion
 * ["marker_0", [3265.59, 853.12]] call zen_area_markers_fnc_updateMarkerPos
```

addons/area_markers/functions/fnc_updateMarkerPos.sqf
```diff
@@ -0,0 +1,21 @@
+#include "script_component.hpp"
```
CreepPork, line 1: Can be removed.

addons/area_markers/functions/fnc_updateAlpha.sqf
```diff
@@ -0,0 +1,34 @@
...
+
+private _localAlpha = if (
+    (playerSide in _sides) ||
+    (!isNull (getAssignedCuratorLogic player)) // ZEUS should always see the markers!
+) then { _alpha } else { 0 };
```
CreepPork, line 32: I don't like this if statement styling, the unnecessary parentheses can be removed in `(playerSide in _sides)` and this should be updated `ZEUS should ...` --> `Zeus should ...`.

addons/area_markers/functions/fnc_configure.sqf
```diff
@@ -93,6 +93,41 @@ private _ctrlAlphaSlider = _ctrlConfigure controlsGroupCtrl IDC_CONFIGURE_ALPHA_
...
+        private _control = _sidesControlGroup controlsGroupCtrl _x;
+        _control ctrlAddEventHandler ["ButtonClick", {
+                params ["_control"];
+                [_control, !(_control getVariable [QGVAR(value), true])] call GVAR(configure_updateSideControl);
+        }];
+        private _controlSide = _foreachindex call bis_fnc_sideType;
+        _control setVariable [QGVAR(side), _controlSide];
+        [_control, _controlSide in _selectedSides] call GVAR(configure_updateSideControl);
+    } forEach IDCS_CONFIGURE_SIDEVISIBILITY_ALL;
+};
```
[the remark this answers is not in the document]
Fusselwurm: Not strictly necessary. I had thought of pulling it in its own function, but didnt follow through 😬

addons/area_markers/functions/fnc_applyProperties.sqf
```diff
@@ -41,6 +41,18 @@ _marker setMarkerColor _color;
...
 
 [QGVAR(updateIcon), [_marker, _rotation, _color]] call CBA_fnc_globalEvent;
+
+private _sidesControlGroup = _ctrlConfigure controlsGroupCtrl IDC_CONFIGURE_SIDEVISIBILITY;
+private _sides = IDCS_CONFIGURE_SIDEVISIBILITY_ALL
+    apply { _sidesControlGroup controlsGroupCtrl _x }
+    apply {
+        if (_x getVariable [QGVAR(value), true]) then {
+             _x getVariable [QGVAR(side), sideUnknown]
+        } else { sideUnknown };
```
CreepPork, line 53: Whoops, one too many spaces.
```suggestion
        if (_x getVariable [QGVAR(value), true]) then {
            _x getVariable [QGVAR(side), sideUnknown]
        } else { sideUnknown };
```

addons/area_markers/functions/fnc_updateAlpha.sqf
Fusselwurm: all of this looks a bit unwieldy.

how about 

```sqf
private _isVisibleSide = playerSide in _sides;
private _isZeus = !isNull (getAssignedCuratorLogic player);
if (_isVisibleSide || _isZeus) then {
    _marker setMarkerAlphaLocal _alpha;
} else {
    _marker setMarkerAlphaLocal 0;
};

```

CreepPork:
```sqf
private _isVisibleSide = playerSide in _sides;
private _isZeus = !isNull (getAssignedCuratorLogic player);

if (_isVisibleSide || _isZeus) then {
    _marker setMarkerAlphaLocal _alpha;
} else {
    _marker setMarkerAlphaLocal 0;
};

**Comment by CreepPork** [comment]
Area marker documentation/images should be updated too.

@mharis001 Should it be handled by this PR or not?

**Comment by mharis001** [comment]
Either works, we can do it later in another PR if that is easier.

**Comment by mharis001** [comment]
This should be changed to use #362's common sides control for the UI handling. Currently, the icons are white when hovered and change side instantly when clicked.

**Comment by mharis001** [comment]
Some things that need to be taken care of for this:
- `playerSide` does not change if the player joins a group of a different side.
- verify that it works as expected for JIP players.
- consider showing all markers to curators when they are in the Zeus interface only, not always.
- fix configure display positioning.

File: addons/area_markers/XEH_postInit.sqf [hunk 12]
```diff
@@ -30,8 +30,17 @@ if (isServer) then {
             [QGVAR(deleteIcon), _marker] call CBA_fnc_globalEvent;
         };
     }] call CBA_fnc_addEventHandler;
+
+    #define SIDES_ARRAY_HASH [[], [east, west, independent, civilian]] call CBA_fnc_hashCreate;
```
CreepPork, line 34: Use new 2.02 hash map

PR #343, opened 2020-03-29
Add careless option to Patrol Area module

### LINKED ISSUE #339: Patrol module should have an option for careless
When placing down civilians you want them not to react to players or other AI. So you would put them into careless and make them patrol. But there is no option in the patrol module so you have to do it separately and that is annoying.

**Solution you'd like:**
In the Patrol module add a new entry in the toolbox for `Careless`.

N/A

**Additional context:**
N/A

SDTheCyanWyan:
- Add a `Careless` option to the Toolbox
- Will close #339

### DIFF
--- a/addons/modules/functions/fnc_modulePatrolArea.sqf
+++ b/addons/modules/functions/fnc_modulePatrolArea.sqf
@@ -49,7 +49,7 @@ if !(alive _unit) exitWith {
     if (_behaviour > 0) then {
         [QEGVAR(common,setFormation), [_unit, "COLUMN"], _unit] call CBA_fnc_targetEvent;
 
-        private _speedMode = ["LIMITED", "NORMAL"] select (_behaviour > 1);
+        private _speedMode = ["LIMITED", "NORMAL"] select (_behaviour > 2);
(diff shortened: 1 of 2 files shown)


addons/modules/functions/fnc_modulePatrolArea.sqf
```diff
@@ -51,7 +52,7 @@ if !(alive _unit) exitWith {
         private _speedMode = ["LIMITED", "NORMAL"] select (_behaviour > 1);
```
mharis001: Should also be changed, not sure what speed mode makes most sense for careless behaviour.

CreepPork: I would assume limited for the most part.

SDTheCyanWyan: Careless behavior is accompanied by Limited speed. Though maybe also having Relaxed be accompanied by Limited speed be a good idea

mharis001: Yes, change it to `_behaviour > 2`.

PR #355, opened 2020-04-06
Use ZEN remote control in ACE interaction menu action

mharis001:
- title, replace ACE using BI's remote control with own

### DIFF
--- a/optionals/compat_ace/functions/fnc_canRemoteControl.sqf
+++ b/optionals/compat_ace/functions/fnc_canRemoteControl.sqf
@@ -1,0 +1,21 @@
+#include "script_component.hpp"
+/*
+ * Author: mharis001
+ * Checks if a unit from the currently selected objects can be remote controlled.
+ *
+ * Arguments:
+ * 0: Curator <OBJECT>
+ *
+ * Return Value:
+ * Can Remote Control <BOOL>
+ *
+ * Example:
+ * [_curator] call zen_compat_ace_fnc_remoteControl
+ *
+ * Public: No
+ */
+
+params ["_curator"];
+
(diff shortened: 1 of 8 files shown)


optionals/compat_ace/functions/fnc_canRemoteControl.sqf
```diff
@@ -0,0 +1,21 @@
...
+ * Example:
+ * [_curator] call zen_compat_ace_fnc_remoteControl
+ *
+ * Public: No
+ */
+
+params ["_curator"];
+
+SELECTED_OBJECTS findIf {_x call EFUNC(remote_control,canControl)} != -1
+&& {[_curator, {"ace_interaction" in curatorAddons _this}, missionNamespace, QACEGVAR(interaction,zeusCheck), 1e9, "ace_interactMenuClosed"] call ACEFUNC(common,cachedCall)}
```
CreepPork, line 21: This whole statement is very hard to understand. Could it be separated into different statements with comments or something to make it less confusing (there's code, namespaces, 1e9 and so on).

mharis001: I can see how on first read-though it looks weird but that is just how the arguments for `cachedCall` are. Can't really be changed.

CreepPork: The arguments could be extracted into separate variables which would try to achieve named parameters.

So you wouldn't need to look up what would be the arguments of that function but just look how the parameters are named.

mharis001: I think that is unnecessary for this situation (or for which argument it is helpful, you just have to know how the function works).

PR #356, opened 2020-04-06
Add new chairs to Sit On Chair module

SDTheCyanWyan:
Add 3 chairs for Sit On Chair module
- Rugged Desk Chair (Black)
- Rugged Desk Chair (Olive)
- Rugged Desk Chair (Sand)


addons/modules/functions/fnc_moduleSitOnChair.sqf
```diff
@@ -45,6 +51,8 @@
     180, \
     0, \
     180, \
+    0, \
```
mharis001, line 54: Adding 4 chair types but only 2 directions.

SDTheCyanWyan: Fixed it. Sorry about that.

PR #363, opened 2020-04-17
Add execution mode button to exec attributes

mharis001:
- Add button to toggle between local, target, and global execution for object and group exec attributes


mharis001:
Changed to use letter images from the base game. These look better and are centered.

PR #364, opened 2020-04-21
Add Rotate Object module

### LINKED ISSUE #104: Feature Request: Rotate objects
Useful module for rotating objects so you don't have to flip them with altitude and perhaps adding local flipping to visualize what it would look like would be incredibly useful but dunno how do-able it is.

mharis001:
- title, Close #104


addons/modules/XEH_postInit.sqf
```diff
@@ -73,3 +73,10 @@ if (isServer) then {
         _unit setVehiclePosition [_position, [], 0, "NONE"];
     }, _this] call CBA_fnc_execNextFrame;
 }] call CBA_fnc_addEventHandler;
+
+[QGVAR(setRotation), {
+    params ["_object", "_pitch", "_roll", "_yaw"];
```
bux, line 78: my OCD says:
"roll" "pitch" "yaw" is the correct order 😄

PR #365, opened 2020-04-22
Create a new component easier via a Python tool

CreepPork:
- Add `new_component.py` script which takes a single argument (component name, e.g. `ai`)
- It will behave weirdly with more than one word so it isn't a catch-all solution (e.g. `area_markers`).
  - You will need to manually change the contents of the `stringtable.xml`  and the beautified name in `script_component.hpp`.
- _Note_: it does a search & replace of the word `blank|Blank|BLANK` and requires a `blank` component in the `extras` directory. If your project doesn't follow those guidelines, it will require manual tweaking.

### DIFF
--- a/tools/new_component.py
+++ b/tools/new_component.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import shutil
+import re
+
+
+def main():
+    root_dir = ".."
+    if os.path.exists("addons"):
+        root_dir = "."
+
+    os.chdir(root_dir)
+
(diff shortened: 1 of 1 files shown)


bux:
Might be a good time to start using hemtt instead of reimplementing its functionality 😄

CreepPork:
HEMTT templating is currently not implemented, once it will be and #55 will be merged this script can be removed.

bux:
Oh it is or rather was. Depends on what version you want to be using. Guess, 0.6.x would work fine too.

PR #132, opened 2019-08-24
Add Throw Grenade context menu action

mharis001:
**When merged this pull request will:**
- Add context menu actions to make AI throw grenades at the selected position
    - Close #117 
    - Close #609 
- Add `isSwimming` common function

<details>
<summary>Image</summary>

throw_grenades_2

</details>

PR #371, opened 2020-05-07
Projectiles

ampersand38:
- Add modules to have AI throw grenades, fire GL, and fire launcher, on high or low ballistic trajectory


synixebrett:
Is it possible to predict the arc? I am not a fan of the module for switching between high / flat. It would be nice to have an approximate arc, and use a keybind to switch modes while placing the target.

addons/modules/XEH_PREP.hpp
```diff
@@ -72,10 +74,20 @@ PREP(moduleSpawnDestroyer);
...
 PREP(moduleToggleIRLasers);
 PREP(moduleToggleLamps);
+PREP(moduleToggleThrowTrajectory);
 PREP(moduleTurretOptics);
 PREP(moduleUnGarrison);
 PREP(moduleVisibility);
 PREP(moduleWeather);
+PREP(projectiles_unit);
+PREP(projectiles_zeus);
+PREP(moduleWeather);
```
synixebrett, line 93: Accidental duplication?

addons/modules/CfgVehicles.hpp
```diff
@@ -300,6 +300,59 @@ class CfgVehicles {
...
+		displayName = CSTRING(moduleThrowSmoke);
+        function = QFUNC(moduleThrowSmoke);
+    };
+    class GVAR(moduleThrowStone): GVAR(moduleBase) {
+        curatorCanAttach = 1;
+        category = QGVAR(Projectiles);
+		displayName = CSTRING(moduleThrowStone);
+        function = QFUNC(moduleThrowStone);
+    };
+    class GVAR(moduleThrowCan): GVAR(moduleBase) {
```
synixebrett, line 332: Personally I don't think these should be in ZEN, but rather ALiVE. They should at least be hidden if ALiVE is not loaded.

addons/modules/functions/fnc_moduleThrowSmoke.sqf
```diff
@@ -0,0 +1,42 @@
+#include "script_component.hpp"
+/*
+ * Author: Ampersand
+ * Zeus module function to make unit throw a smoke grenade.
```
synixebrett, line 4: Perhaps create a dialog with a dropdown for the colors of smoke the unit has? Rather than only white.

addons/modules/functions/fnc_projectiles_unit.sqf
```diff
@@ -0,0 +1,174 @@
...
+	_unit removeEventHandler ["Fired", _thisEventHandler];
+	_unit setVariable ["zen_projectiles_thrown", true];
+	_unit enableAI "PATH";
+	
+}];
+
+_unit setVariable ["zen_projectiles_thrown", false];
+_unit setVariable ["zen_projectiles_time", diag_tickTime];
+
+_this spawn {
```
synixebrett, line 124: Use of spawn and waitUntil, could this not be achieved with CBA instead?

jonpas:
How does one select a grenade that's not part of that selection? Eg. some mod adds a dart _(just a random thing I came up with right now)_, with this implementation of multiple modules, can one do that?

That arc doesn't really seem to be an arc, it's just a straight line? I don't really see the reason for that as a module, could be a CBA Setting for those scenarios where Zeus needs to be limited or someone can't handle line drawing performance. No need to clutter the modules.

synixebrett:
> How does one select a grenade that's not part of that selection?

A module for just throwing that gets all the throwables would be better, could be done with [BIS_fnc_isThrowable](https://community.bistudio.com/wiki/BIS_fnc_isThrowable)

addons/modules/XEH_PREP.hpp
ampersand38: Yes, sorry.

ampersand38:
> Is it possible to predict the arc? I am not a fan of the module for switching between high / flat. It would be nice to have an approximate arc, and use a keybind to switch modes while placing the target.

It's using the `initSpeed` can calculating the ballistic arc to hit the targetPos. There's always 2 possible arcs.

ampersand38:
> That arc doesn't really seem to be an arc, it's just a straight line? I don't really see the reason for that as a module, could be a CBA Setting for those scenarios where Zeus needs to be limited or someone can't handle line drawing performance. No need to clutter the modules.

High Arc would be needed for hitting a roof top, for example.

ampersand38:
> > How does one select a grenade that's not part of that selection?
> 
> A module for just throwing that gets all the throwables would be better, could be done with [BIS_fnc_isThrowable](https://community.bistudio.com/wiki/BIS_fnc_isThrowable)

That would be a useful feature. I've worked on the standard stuff because it's faster to not have to choose a throwable.

jonpas:
What about using ZENs context menu? Select AI, right click at location to throw. Clicking "Throw" would throw something based on some priority, sub-menu for manual selection.

addons/modules/CfgVehicles.hpp
ampersand38: I could remove those until I figure out how to do that.

ampersand38:
> What about using ZENs context menu? Select AI, right click at location to throw. Clicking "Throw" would throw something based on some priority (or just first available), sub-menu for manual selection.
> 
> Same for "Launch", launcher or GL.

Is it currently possible to open context menu with AI selected? The default right click of course sets a waypoint.

addons/modules/functions/fnc_moduleThrowSmoke.sqf
ampersand38: The current modules were created with the focus of "here's what the players need to see". I'll be creating another module where Zeus can choose from the magazines available on the AI unit to cover this functionality.

bux:
> Are grenade commands used frequently enough to deserve a spot on every unit's context menu?

imho, no

ampersand38:
Latest Throw Select module lists unit's available throwables in context menu between dropping the module and selecting target position.

I haven't worked out how to remove the actions if menu is closed without choosing one of the throwables. https://youtu.be/HItqwS8pbkk

bux:
>Are grenade commands used frequently enough to deserve a spot on every unit's context menu?

Possible way to fix this could be the following idea, that is - of course - not in scope of this PR.
Use keyboard modifiers to open different context menus:

- Hold down <kbd>Alt</kbd> and right click to get the offensive targeted context menu actions like throwing grenades or a quick access to a "search and destroy" waypoint
- Hold down <kbd>Shift</kbd> while right clicking to get the more defensive targeted context menu actions. ....
- etc ...

I don't know if that's possible at all, maybe that's an idea @mharis001 could use.

Such modifiers to open different context menus are already used in different graphic software. So it's not something completely uncommon 😄

addons/modules/functions/fnc_projectiles_unit.sqf
ampersand38: New implementation with CBA.

CreepPork:
I would let you know a similar feature is in development too: #132

ampersand38:
`spawn` and `waitUntil` have been CBA-ized. 
Throw selected grenade is working properly.
Fire Launcher is more reliable.

mharis001:
I dislike the addition of this many modules. I think it can be limited to Fire Launcher, Fire GL, and Throw Grenade. Modules that require ALiVE should be removed. Also, the various actions do not work if the unit is in combat.

The current implementation into the context menu is pretty rough. If we want to keep that, it would be better to have only grenade actions shown. That requires a small change to `context_menu` which would allow entirely different action trees to be used when opening the menu - something I was planning on doing eventually for #132 .

For what @bux suggested, `cba_events_shift` and `cba_events_alt` can be used for the action conditions to achieve that.

ampersand38:
I've removed and renamed the extraneous modules. I've noticed that throw commands sent while units are in certain actions will be delayed or ignored, but most are executed. https://youtu.be/xOgiUzg8214

ampersand38:
Added icons https://steamuserimages-a.akamaihd.net/ugc/1026203216031872831/AF6C7BF14D3E4929B38548A265ED14D64A02DA35/

oOKexOo:
I really like these features. I considered getting launchers working for AI as impossible for some while. It works as good as it can considering how hard it is getting alerted ArmA AI to do what you want them to do. I have a few comments:
- Did you try get missile launchers working. There apparently is a `setMissileTarget` command. I have never used it so far tbh.
- It seems that I somehow managed to break the modules when I used them a lot. Unfortunately, I haven't figured it out how to reproduce it. When it is broken, the module is always executed for the same unit, no matter on which unit I place the module.
- Throw select: It would be better to get rid of all the other context menu entries that are unrelated to the module.

oOKexOo:
I'm curious about your formula for calculating the initial angle. Do you have a reference or derivation for it?

ampersand38:
> I'm curious about your formula for calculating the initial angle. Do you have a reference or derivation for it?

I got it from this video https://www.youtube.com/watch?v=bqYtNrhdDAY

ampersand38:
Now working with soft launch weapons (RPG, NLAW, Titan, Vorona) https://youtu.be/gSjC2l98-Ps

CreepPork:
For the grenades a graphical drawing like ACE's advanced throwing would be excellent for this.

PR #374, opened 2020-05-11
Rework and improve dialog component

### LINKED ISSUE #335: Select all / deselect all options for the Teleporter module
When teleporting players around from respawn, I have to go into and deselect multiple entries and then select those that need to be teleported back.
Sometimes, I may miss somebody and teleport them back which is quite bad.

**Solution you'd like:**
The Teleport module should have two buttons somewhere at the top, that selects all the units in the list and deselects.

N/A

**Additional context:**
N/A

mharis001:
- Make various improvements to the `dialog`  component
- Refactor how content controls return their values
    - Rather keep track of the value as changes are made, controls define a `QFUNC(value)` function that returns their current value
    - Simplifies a lot of the control specific functions since a lot less handling is needed
- Use common base UI classes and display for the dialog
    - Fix scrollbar overlap with content controls when the content height is greater than the maximum
- Add `_saveID` parameter to `create` function specify a custom save ID
    - Needed when, for example, the title may change between dialog multiple openings of the dialog but the content remains the same
- Add `_args` parameter to `initSliderEdit` common function
- Fix `initDisplayPositioning` common function not setting the content's height when it is greater than the maximum
- Fix issue where the saved value was modified even though the dialog was cancelled for `COLOR`, `OWNERS`, and `VECTOR` control types
    - Saved value array was being modified by reference

**Control specific changes:**
- `COLOR`
    - Changed the color preview to be bigger
    - Changed to use common `initSliderEdit` for the UI handling
- `COMBO` / `LIST`
    - Added option to sort entries alphabetically
    - `LIST` now uses `COMBO`'s function since handling is identical other than setting the list height
- `OWNERS`
    - Moved UI to common so it can be used by other components
    - Added `initOwnersControl` common function to setup its handling
    - Added buttons to (un-)check all groups/players, Close #335 
- `SIDES`
    - Changed to use common sides control
    - Added ability to have multi-select sides controls, specified by an array of sides as the default value
- `SLIDER`
    - Changed to use common `initSliderEdit` for the UI handling


mharis001:
Will update documentation in another PR.

PR #376, opened 2020-05-13
Improve Game Master module initialization

### LINKED ISSUE #113: Zeus Interface gets stuck in infinite loop trying to open when respawnOnStart = 1
This occurs in vanilla Zeus as well, and I'm doubtful it can be fixed but I'm creating an issue here to maybe start work on a fix, if possible.

The issue leaves the Curator in an infinite loop of the Curator interface trying to start but it then closes and the map (with the position/loadout options selected) appears but then quickly disappears. And this behaviour continues until the Curator is forced to exit the game via ALT-F4 or in some other way ending the process.

Perhaps a Feedback tracker ticket could be created as well.

### LINKED ISSUE #372: Double clicking a unit opens 2 menus
**Arma 3 Version:** `x.xx` stable
**CBA Version:** `3.x.x` stable 
**ZEN Version:** `1.x.x` stable
**Mods:**
```
- CBA_A3
- ZEN
-armazuescache
```

**Description:**
- Add a detailed description of the error. This makes it easier for us to fix the issue.

**Steps to reproduce:**
- Add the steps needed to reproduce the issue.
when given curator or even logging into the server as admin to access zues, you double click a unit or object and the window that appears, instead of one 2 show up. so when you hit ok on one, the other one is there with default information, they seem to work either one you use, but it anoying to have to close 2 and sometimes i can mess up what objective or setting you have put into it
- Dedicated

**Additional information:**
- Provide any additional information that will help us solve this issue.

**RPT log file:**
- Add a link ([gist](https://gist.github.com) or [pastebin](http://pastebin.com)) to the client and/or server RPT file. An instruction to find your RPT files can be found [here](https://community.bistudio.com/wiki/Crash_Files#Arma_3).

**Screenshots:**

mharis001:
- Replace `ModuleCurator_F`'s initialization function with our own
    - Follows BI's function but code has been cleaned up and modernized
    - Fix double attributes dialog bug for Game Master modules created after mission init, Close #372 
    - Fix respawn on start with forced interface preventing interaction with the respawn menu, Close #113 
    - Fix loss of curator assignment when the module's locality is transferred, Caused the bug where the player is forced to respawn in order to access Zeus
    - Unable to get rid of `activateAddons` code even though CBA activates all addons, did not work in singleplayer
- Add settings to control ascension messages and camera eagle, like ACE
- Add `runAfterSettingsInit` common helper function


bux:
Maybe add a setting to disable Zeus pings?

Would remove the need to set `missionNamespace setvariable ["bis_fnc_curatorPinged_time", 99999999, true];` in missions

mharis001:
The player's icon (in Zeus view) would still have the grow effect since that is handled in engine. The sound and icon effect (white Zeus logo in screen center) for the player can be disabled though.

addons/modules/functions/fnc_bi_moduleCurator.sqf
```diff
@@ -0,0 +1,397 @@
...
+if (_activated) then {
+    // BI's module framework is bugged and calls module function twice on clients after mission start
+    if (_logic getVariable [QGVAR(initialized), false]) exitWith {};
+
+    // Mark the module as initialized on this machine, fixes double attributes dialog bug
+    _logic setVariable [QGVAR(initialized), true];
+
+    // Exit if the module is not created on the server
+    if (!isServer && {local _logic} && {isNull getAssignedCuratorUnit _logic}) exitwith {
+        [format ["%1 is trying to create a Curator logic ModuleCurator_F.", profileName], "BIS_fnc_error", false] call BIS_fnc_MP;
```
veteran29, line 34: This is the same as:
```suggestion
        [format ["%1 is trying to create a Curator logic ModuleCurator_F.", profileName] remoteExec ["BIS_fnc_error", 2];
```
(it spawns the function like original, use `remoteExecCall` for unscheduled.)

addons/modules/functions/fnc_bi_moduleCurator.sqf
```diff
@@ -0,0 +1,397 @@
...
+
+    // Wipe out the variable so clients can't access it
+    _logic setVariable ["owner", nil];
+
+    if (isServer) then {
+        // Prepare admin variable
+        private _adminVar = "";
+
+        if (_isAdmin) then {
+            private _letters = ["a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z"];
```
CreepPork, line 70: Maybe to a define? To keep it a bit cleaner.

addons/modules/functions/fnc_bi_moduleCurator.sqf
```diff
@@ -0,0 +1,397 @@
...
+            } forEach synchronizedObjects _logic;
+
+            _addons call BIS_fnc_activateAddons;
+        };
+    };
+
+    if (hasInterface) then {
+        waitUntil {local player};
+
+        private _serverCommand = if (_ownerVar == "#adminLogged") then {"#shutdown"} else {"#kick"};
```
CreepPork, line 288: This to a select

addons/modules/functions/fnc_bi_moduleCurator.sqf
mharis001: Don't think it will make a difference, this is BI code anyways.

PR #379, opened 2020-05-14
Add Orient to Terrain Normal keybind

mharis001:
- title, complementary to the base game one to orient to sea normal


addons/editor/initKeybinds.sqf
```diff
@@ -57,6 +57,18 @@
     };
 }, {}, [DIK_F, [false, true, true]]] call CBA_fnc_addKeybind; // Default: CTRL + ALT + F
 
+[ELSTRING(main,DisplayName), QGVAR(orientTerrainNormal), "str_3den_display3den_entitymenu_setatl_text", {
```
CreepPork, line 60: Maybe a tooltip if that is possible for those who don't know what that means?

PR #384, opened 2020-05-21
Unlock malaria-infected civs

oOKexOo:
#### Preface:
The Old Man update introduced malaria-infected civilians, however these classes are currently not visible in Eden and Zeus. The way they work is that they change the wound textures of the base civilian classes, which means that the infection is only visible when the unit is injured.

#### When merged this pull request will:
- Unlock malaria-infected civilian classes.
- Set the proper identities for malaria-infected civilian (fixes the faces).
- Add damage to malaria-infected civilian to make the infection visible.

### DIFF
--- a/addons/characters/CfgEditorSubcategories.hpp
+++ b/addons/characters/CfgEditorSubcategories.hpp
@@ -1,0 +1,5 @@
+class CfgEditorSubcategories {
+    class EdSubcat_Personnel_Sick {
+        displayName = CSTRING(MenSick);
+    };
+};
--- a/addons/characters/CfgVehicles.hpp
+++ b/addons/characters/CfgVehicles.hpp
@@ -1,0 +1,95 @@
+class CfgVehicles {
+    // External class reference
+	class C_Man_casual_1_F_afro;
+	class C_Man_casual_3_F_afro;
+	class C_Man_casual_4_F_afro;
+	class C_Man_casual_5_F_afro;
+	class C_Man_casual_6_F_afro;
+	class C_man_polo_1_F_afro;
+	class C_man_polo_2_F_afro;
+	class C_man_polo_3_F_afro;
+	class C_man_polo_6_F_afro;
+	class C_man_sport_2_F_afro;
+
+    class C_Man_casual_1_F_afro_sick : C_Man_casual_1_F_afro {
+		editorSubcategory = "EdSubcat_Personnel_Sick";
+        scope = 2;
+        scopeCurator = 2;
+        class EventHandlers {
+            init = "(_this select 0) setIdentity selectRandom ['BIS_Ambient01_sick','BIS_Ambient02_sick','BIS_Ambient03_sick','BIS_Arthur_Sick','BIS_Howard_sick','BIS_John_sick','BIS_Lucas_sick','BIS_Renly_sick']; (_this select 0) setDamage 0.45";
+        };
+    };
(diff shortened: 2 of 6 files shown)


addons/characters/config.cpp
```diff
@@ -0,0 +1,21 @@
...
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"zen_common","a3_characters_f_oldman"};
+        author = ECSTRING(main,Author);
+        authors[] = {"Kex"};
+        url = ECSTRING(main,URL);
+        VERSION_CONFIG;
+
+        // this prevents any patched class from requiring this addon
```
CreepPork, line 15:
```suggestion
        // This prevents any patched class from requiring this addon
```

addons/characters/CfgVehicles.hpp
```diff
@@ -0,0 +1,104 @@
...
+    class C_man_polo_3_F_afro;
+    class C_man_polo_6_F_afro;
+    class C_man_sport_2_F_afro;
+
+    class C_Man_casual_1_F_afro_sick : C_Man_casual_1_F_afro {
+        editorSubcategory = "EdSubcat_Personnel_Sick";
+        scope = 2;
+        scopeCurator = 2;
+        class EventHandlers {
+            init = "(_this select 0) setIdentity selectRandom ['BIS_Ambient01_sick','BIS_Ambient02_sick','BIS_Ambient03_sick','BIS_Arthur_Sick','BIS_Howard_sick','BIS_John_sick','BIS_Lucas_sick','BIS_Renly_sick']; (_this select 0) setDamage 0.45";
```
CreepPork, line 19: Couldn't all of this be extracted to a define?

oOKexOo: If only multiple inheritance was a thing...
Do you suggest to move the entire block to a define or just the init line?

CreepPork: It doesn't look like they change so maybe all the block? And shouldn't this init line be macro'd?

bux:
Is Zen the right mod for this?
CBA or ACE would make more sense, wouldn't it?

oOKexOo:
It's pretty much a modification the get them working in the editor. These classes can already be used without any config modifications in scripts. That's how Old Man does it. However, we could consider moving it to CBA.

addons/characters/CfgEditorSubcategories.hpp
```diff
@@ -0,0 +1,5 @@
+class CfgEditorSubcategories {
+    class EdSubcat_Personnel_Sick {
```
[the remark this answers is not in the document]
CreepPork: I’m as well not sure on this but maybe sometime in the future some other mod would create such a class too and we’d have a conflict. So perhaps using the GVAR might be better and it really doesn’t change anything that much.

CreepPork:
@bux

It could be debated that the unlocking of music classes for Zeus could also be moved to CBA, but at the same time it could be left in as it is regarding the Zeus.

neilzar:
I would say this is better in CBA. It also applies to the editor, so already goes beyond Zeus. It also doesn't require any ZEN functions so there is no real reason to prefer it here over CBA. I say CBA over ACE because music is already unlocked in CBA as well.

bux:
@CreepPork true
Move it to CBA then 👍 😄

oOKexOo:
Closed in favor of CBATeam/CBA_A3#1344

PR #393, opened 2020-05-23
CBRN hose support

Kexanone:
- connect CBRN hoses for exported SQF, vanilla arsenal and deserialized units.
- not fix it for ACE arsenal. This should be done on ACE's side.


addons/common/functions/fnc_openArsenal.sqf
```diff
@@ -23,5 +23,18 @@ if (GVAR(preferredArsenal) == 1 && {isClass (configFile >> "CfgPatches" >> "ace_
 
     [_unit, _unit, true] call ace_arsenal_fnc_openBox;
 } else {
+    GVAR(arsenalCenter) = _unit;
+
+    // Connect CBRN hoses when arsenal is closed
+    if (isNil QGVAR(arsenalCbrnEH)) then {
+        GVAR(arsenalCbrnEH) = [
+            missionnamespace,
```
CreepPork, line 31: missionNamespace

addons/common/functions/fnc_exportMissionSQF.sqf
```diff
@@ -114,6 +114,7 @@ private _fnc_processInventory = {
 
     if (_object isKindOf "CAManBase") then {
         _outputObjects pushBack ["%1 setUnitLoadout %2;", _varName, getUnitLoadout _object];
+        _outputObjects pushBack ["%1 call BIN_fnc_CBRNHoseInit;", _varName];
```
neilzar, line 117: The incorrect facemask is returned by `getUnitLoadout`, leaving the unit without a respirator when the SQF is executed but the hose still exists.

neilzar: Maybe a frame delay will fix this issue?

neilzar: Checking again the mask appears for one frame and is then removed for some reason.

oOKexOo: Yes, `BIN_fnc_CBRNHoseInit` and `setUnitLoadout` needs to be executed in the next frame. Dunno how I missed it.

oOKexOo: Should be resolved now

neilzar (approved):

addons/common/functions/fnc_exportMissionSQF.sqf
```diff
@@ -110,11 +110,17 @@ private _fnc_processGroup = {
...
-
     if (_object isKindOf "CAManBase") then {
-        _outputObjects pushBack ["%1 setUnitLoadout %2;", _varName, getUnitLoadout _object];
+        _outputObjects pushBack "[{";
+        _outputObjects pushBack "    params [""_unit""];";
+        if !(_object call FUNC(hasDefaultInventory)) then {
+            _outputObjects pushBack ["    _unit setUnitLoadout %1;", getUnitLoadout _object];
+        };
+        _outputObjects pushBack "    _unit call BIN_fnc_CBRNHoseInit;";
+        _outputObjects pushBack ["}, [%1]] call CBA_fnc_execNextFrame;", _varName];
```
mharis001, line 120: I think we should decide whether we want export mission SQF to require any mods (CBA in this case) to work. If so, we could then change portions of the code to just call ZEN function as well.

Kexanone: I forgot about that. I can easily swap it with `BIS_fnc_addStackedEventHandler`.

CreepPork, line 120: What's with the spaces?

Kexanone: Proper indentation?

Issue #270, opened 2019-12-09
When Zeus module is set to "Addons -> None" right click menu still displays


bux:
**Arma 3 Version:** `1.96`
**CBA Version:** `3.13.0`
**ZEN Version:** `1.6.0`

**Mods:**
```
- CBA_A3
- ZEN
- ACE3
- BWMod
- CUP (Units, Vehicles, Weapons)
```

**Description:**
- When setting Zeus Game Master Module to not use addons the right click menu still shows thus circumventing the restriction placed on the game master

**Where did the issue occur?**
- Dedicated

**Additional information:**
- Occured in KP_Liberation mission
https://github.com/KillahPotatoes/KP-Liberation

mharis001:
You can force the `zen_context_menu_enabled` setting to `0` to disable the context menu.

PR #419, opened 2020-06-28
Preload Addons

mharis001:
- Add `CfgAddons` entries for all components
- Adds ZEN addons to curators even if "All addons (including unofficial ones)" isn't selected in Game Master module
- Should help with `curatorAddons` being lost after game save/load

### DIFF
--- a/addons/ai/config.cpp
+++ b/addons/ai/config.cpp
@@ -14,4 +14,6 @@ class CfgPatches {
     };
 };
 
+PRELOAD_ADDONS;
+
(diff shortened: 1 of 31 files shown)


bux:
You mean like in #270

CreepPork:
Yeah, something like that.

mharis001:
Curator addons can still be managed using script commands or by the "None" option in the Game Master module. This makes ZEN addons act like "official" base game ones in that they are added by default.

PR #427, opened 2020-07-08
Overhaul Custom Objective module

### LINKED ISSUE #78: Feature Request: Custom Objective Icons
As much as @CreepPork might gut me for this... 😆 

It would be nice to be able to use the task icons that **Eden** has in the **Custom Objective** module in Zeus. It would help the game master better emphasize their tasks and allow players to much more easily understand their task at a glimpse.

customobjectiveicons

mharis001:
- Replace BI's Custom Objective module with our own
- No longer uses FSM for the task handling
- Can make use of ZEN's improved owners control (select multiple groups/players)
- Can select task type/icon, Close #78 
- Add stringtable entries for task types

<details>
<summary>Image</summary>

task_module

</details>


addons/tasks/functions/fnc_update.sqf
```diff
@@ -0,0 +1,77 @@
...
+};
+
+private _newOwners = _owners select (_owners select 3);
+private _oldOwners = _logic getVariable [QGVAR(owners), []];
+
+// Use the module as the task's destination if needed
+private _target = [objNull, _logic] select _showDestination;
+
+// Create/update the task with the new paramters
+[_taskID, _newOwners, [_description, _title, ""], [_target, true], _state, nil, nil, nil, _type] call BIS_fnc_setTask;
```
CreepPork, line 60: Maybe into a new line all arguments?

mharis001:
Used "Task Type" since that is what the 3DEN module has. As for the display names, they are all lowercase and un-localized in the config. Only thing we could do about that is replace the `displayName` entries in `CfgTaskTypes` with our own.

CreepPork:
I'm more thinking that we could just run `CBA_fnc_capitalize` on each of those task types (something like `apply` or something).

mharis001:
I tried `CBA_fnc_capitalize` and it works well for most of the names but we still have things such as "Getin" or "Move4".

CreepPork:
Those are just a few exceptions, maybe we could just replace those or just leave it with `capitalize`. I don't mind, but way better that everything lowercase.

mharis001:
Went ahead and added entries for all of the task types. Seems like the best solution.

addons/tasks/functions/fnc_update.sqf
mharis001: A bit long but I think this is easier to read as a single line.

PR #437, opened 2020-07-14
Add Open Attributes Display keybind

ampersand38:
- Add keybind [`] accent grave to open the Attributes dialog for the selected entity
- It can sometimes be difficult to doubleclick on a moving entity or to find it in the list


addons/attributes/functions/fnc_openForSelection.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * Example:
+ * [] call zen_attributes_fnc_openForSelection
+ *
+ * Public: No
+ */
+
+curatorSelected params ["_objects", "_groups", "_waypoints", "_markers"];
+
+switch (true) do {
+    case !(_groups isEqualTo []): {[_groups select 0, "Group"] call zen_attributes_fnc_open};
```
mharis001, line 21: Call `BIS_fnc_showCuratorAttributes` to open attributes - it handles not opening attributes for objects that have them disabled and opening `curatorInfoType` dialog for objects that still use that.

addons/editor/initKeybinds.sqf
```diff
@@ -82,3 +82,9 @@
         };
     };
 }, {}, [0, [false, false, false]]] call CBA_fnc_addKeybind; // Default: Unbound
+
+[ELSTRING(main,DisplayName), QGVAR(attributesDisplay), [LSTRING(AttributesDisplay), LSTRING(AttributesDisplay_Description)], {
+    if (!isNull curatorCamera && {!dialog && {!GETMVAR(RscDisplayCurator_search,false)}}) then {
+        [] call EFUNC(attributes,openForSelection);
+    };
+}, {}, [DIK_GRAVE, [false, false, false]]] call CBA_fnc_addKeybind; // Default: ` (Accect Grave / Tilde)
```
mharis001, line 90:
```suggestion
}, {}, [DIK_GRAVE, [false, false, false]]] call CBA_fnc_addKeybind; // Default: GRAVE
```

addons/editor/initKeybinds.sqf
```diff
@@ -82,3 +82,9 @@
         };
     };
 }, {}, [0, [false, false, false]]] call CBA_fnc_addKeybind; // Default: Unbound
+
+[ELSTRING(main,DisplayName), QGVAR(attributesDisplay), [LSTRING(AttributesDisplay), LSTRING(AttributesDisplay_Description)], {
```
mharis001, line 86: Move this to the `attributes` component.

addons/editor/stringtable.xml
```diff
@@ -210,5 +210,11 @@
             <Japanese>全編集可アイコンの視認性を切り替えます。一時的な処理軽減や整理に使用できます。</Japanese>
             <Polish>Przełącza widoczność wszsytkich ustawionych ikon, przydatne by tymczasowo zmniejszyć lag, bądź niepotrzebne zaśmiecenie ekranu gdy znajduje się na nim dużo ikon.</Polish>
         </Key>
+        <Key ID="STR_ZEN_Editor_AttributesDisplay">
+            <English>Show Attributes Display</English>
```
mharis001, line 214: I think "Open" is better than "Show" for this.

addons/attributes/functions/fnc_openForSelection.sqf
ampersand38: Ahh, I figured out it's `_entity call BIS_fnc_showCuratorAttributes`. I was getting error with `[_entity] call BIS_fnc_showCuratorAttributes` and thought it wasn't used or something.

addons/attributes/initKeybinds.sqf
```diff
@@ -0,0 +1,5 @@
+[ELSTRING(main,DisplayName), QGVAR(attributesDisplay), [LSTRING(AttributesDisplay), LSTRING(AttributesDisplay_Description)], {
```
mharis001, line 1:
```suggestion
[ELSTRING(main,DisplayName), QGVAR(open), [LSTRING(OpenAttributesDisplay), LSTRING(OpenAttributesDisplay_Description)], {
```

addons/attributes/initKeybinds.sqf
```diff
@@ -0,0 +1,5 @@
...
+    if (!isNull curatorCamera && {!dialog && {!GETMVAR(RscDisplayCurator_search,false)}}) then {
+        [] call EFUNC(attributes,openForSelection);
```
mharis001, line 3: I'd say move all keybind handling code into this file - generally how it is done in other components.

ampersand38: Do you mean move fnc_openForSelection into the code for this keybind?

ampersand38:
In Zeus, with a single unit in its own group, it's possible to select either the unit or the group, shown by different highlighting in the Edit entities list. I'll try to find out if it's possible to distinguish the highlighted entity.

addons/attributes/initKeybinds.sqf
mharis001: Yes.

mharis001:
I think the highlighting differences result from which item has been most recently selected in multi-select `CT_TREE` controls.

ampersand38:
I'm pretty lost on investigating how to pull information out of the tree display.

mharis001:
There isn't a good way. For hovered entities, you can use the `curatorMouseOver` command.

ampersand38:
Found a solution using `Curator______SelectionChanged` EHs.

mharis001:
Now, you can open attributes of a group even after it has been deselected (select a group, click ground to deselect, use keybind). I think allowing a specific entity to be specified by hovering over it will be the simplest solution.

ampersand38:
That would be straight forward, I agree. However, sometimes with moving units on the main view and ALiVE spawning stuff to shift the list view it's inconvenient to mouse over.

addons/attributes/XEH_preInit.sqf
```diff
@@ -6,17 +6,45 @@ PREP_RECOMPILE_START;
...
+GVAR(selectedEntity) = objNull;
+
 ["ModuleCurator_F", "Init", {
     params ["_logic"];
 
     _logic addEventHandler ["CuratorMarkerPlaced", {call FUNC(handleMarkerPlaced)}];
+
+    // handle group being select when all of its sub entities are selected
+    private _fnc_checkGroupSelection = {
+        private _entity = _this # 1;
```
CreepPork, line 28: Afaik, we don't use `#`.

addons/attributes/XEH_preInit.sqf
```diff
@@ -6,17 +6,45 @@ PREP_RECOMPILE_START;
...
 
+GVAR(selectedGroup) = grpNull;
+GVAR(selectedEntity) = objNull;
+
 ["ModuleCurator_F", "Init", {
     params ["_logic"];
 
     _logic addEventHandler ["CuratorMarkerPlaced", {call FUNC(handleMarkerPlaced)}];
+
+    // handle group being select when all of its sub entities are selected
```
CreepPork, line 26:
```suggestion
    // Handle group being select when all of its sub entities are selected
```
Do this for all comments

CreepPork:
Would it be possible to open the context menu where it only displays the crew members or units in a group? You could select the unit you want that way I guess.

addons/attributes/XEH_preInit.sqf
ampersand38: https://github.com/ampersand38/ZEN/commit/35b77767cd0d95abd8411a11155d093cd301648c

ampersand38:
Right now it's working for the last selected single entity.
Do you see a use case for selecting a number of entities and then:
- selecting one via context menu
- cycling through each one's attributes dialog with repeated key presses

CreepPork:
Cycling would be time consuming and it would be hard to differentiate between all the units because it would only show the unit’s name. 

In the context menu you could show the unit’s icon (medic, AT), name, and maybe some other attribute.

addons/attributes/XEH_preInit.sqf
CreepPork: You don’t need to attach these commit info, just mark it solved otherwise a lot of spam emails :)

ampersand38:
I mean, there's already the entities list and the main view for making a selection. This PR was meant to resolve the issue of double clicking on a UI element that is moving, either due to unit movement or spawn/deletion. I think trying to make another layer of entity selection would be out of scope.

addons/attributes/stringtable.xml
```diff
@@ -238,5 +238,11 @@
             <English>Abilities</English>
             <Polish>Zdolności</Polish>
         </Key>
+        <Key ID="STR_ZEN_Attributes_OpenAttributesDisplay">
+            <English>Open Attributes Display</English>
+        </Key>
+        <Key ID="STR_ZEN_Attributes_OpenAttributesDisplay_Description">
+            <English>Opens the Zeus attribute display for the mouse-overed or selected entity.</English>
```
mharis001, line 245: "attributes" to match the display name.

ampersand38: https://github.com/ampersand38/ZEN/commit/d557e41e9009ab0fdff9e854fc4fccd299171685

ampersand38:
> You can now open attributes for an entity that is hovered but not selected. This should not be allowed since most attributes rely on the entities that they modify to be selected. The suggestion was to allow the user to hover of a specific selected entity to specify which entity to open the attributes for. And if nothing is hovered, follow the priority that you originally had.

I see! That makes sense. I've changed it to only open for mousedOver entity that is already selected.

> I don't think the selection changed handling will be necessary with the behaviour described above.

When a group is selected, all of its sub-entities are also selected. So if the group attributes display is the goal, then curator must select group *and* mouse over the group. Is that acceptable and/or preferable?

mharis001:
> When a group is selected, all of its sub-entities are also selected. So if the group attributes display is the goal, then curator must select group and mouse over the group. Is that acceptable and/or preferable?

Yes, I think it is acceptable to require hovering over group icons.

addons/attributes/initKeybinds.sqf
```diff
@@ -0,0 +1,34 @@
...
+                GVAR(selectedGroup)
+            };
+            case !(_objects isEqualTo []): {_objects select 0};
+            case !(_waypoints isEqualTo []): {_waypoints select 0};
+            case !(_markers isEqualTo []): {_markers select 0};
+            default { nil };
+        };
+        if !(isNil "_entity") then {
+            _entity call BIS_fnc_showCuratorAttributes;
+        };
```
mharis001, line 32: I think this is better written as:

```sqf
{
    private _entity = _x param [0];

    if (!isNil "_entity") exitWith {
        _entity call BIS_fnc_showCuratorAttributes;
    };
} forEach [SELECTED_GROUPS, SELECTED_OBJECTS, SELECTED_WAYPOINTS, SELECTED_MARKERS];
```

addons/attributes/initKeybinds.sqf
```diff
@@ -0,0 +1,34 @@
...
+                };
+                case (_entity in _objects);
+                case (_entity in _groups);
+                case (_entity in _markers): {_entity};
+                default {nil};
+            };
+        };
+        if (!isNil "_mouseOver") exitWith {
+            _mouseOver call BIS_fnc_showCuratorAttributes;
+        };
```
mharis001, line 19: I think this is better written as:

```sqf
scopeName "Main";

curatorMouseOver params ["_type", "_entity", "_index"];

if (_type != "") then {
    if (_type == "ARRAY") then {
        _entity = [_entity, _index];
    };

    if (curatorSelected findIf {_entity in _x} != -1) then {
        _entity call BIS_fnc_showCuratorAttributes;
        breakOut "Main";
    };
};
```

addons/attributes/stringtable.xml
```diff
@@ -238,5 +238,11 @@
             <English>Abilities</English>
             <Polish>Zdolności</Polish>
         </Key>
+        <Key ID="STR_ZEN_Attributes_OpenAttributesDisplay">
+            <English>Open Attributes Display</English>
+        </Key>
+        <Key ID="STR_ZEN_Attributes_OpenAttributesDisplay_Description">
+            <English>Opens the Zeus attributes display for the mouse-overed or selected entity.</English>
```
mharis001, line 245: Should add how the hovering behaviour works. Maybe something like: "... display for the selected entities.\nHovering over a selected entity can be used to specify the exact entity to open the display for."

ampersand38: How's this?
```
    <Key ID="STR_ZEN_Attributes_OpenAttributesDisplay_Description">
            <English>Opens the Zeus attributes display for the selected entity.\nWhen multiple entitie are selected (i.e. due to selecting a Group), the moused-over entity will be used.</English>
        </Key>
```

addons/attributes/initKeybinds.sqf
ampersand38: https://github.com/zen-mod/ZEN/pull/437/commits/e37d2ec65ca358f42f5f0c61e689bb5b9a66152b

addons/attributes/initKeybinds.sqf
ampersand38: https://github.com/zen-mod/ZEN/pull/437/commits/e37d2ec65ca358f42f5f0c61e689bb5b9a66152b

addons/attributes/stringtable.xml
mharis001: Looks good. Should lowercase "group" and maybe use "hovered" instead of "mouse-over".

ampersand38: https://github.com/zen-mod/ZEN/pull/437/commits/1a64b2ab08138927e00752034561b559e635ee1d

addons/attributes/stringtable.xml
```diff
@@ -238,5 +238,11 @@
             <English>Abilities</English>
             <Polish>Zdolności</Polish>
         </Key>
+        <Key ID="STR_ZEN_Attributes_OpenAttributesDisplay">
+            <English>Open Attributes Display</English>
+        </Key>
+        <Key ID="STR_ZEN_Attributes_OpenAttributesDisplay_Description">
+            <English>Opens the Zeus attributes display for the selected entity.\nWhen multiple entitie are selected (i.e. due to selecting a group), the hovered entity will be used.</English>
```
CreepPork, line 245:
```suggestion
            <English>Opens the Zeus attributes display for the selected entity.\nWhen multiple entities are selected (i.e. due to selecting a group), the hovered entity will be used.</English>
```

CreepPork (changes requested):

CreepPork (approved):

PR #481, opened 2020-10-03
Allow opening context menu by right clicking when AI are selected

mharis001:
- Allow opening context menu by right clicking when waypoint placement would normally occur due to selected AI and/or waypoints
    - Enabled through "Override Waypoint Placement" setting (default: `false`)
    - Can use <kbd>SHIFT</kbd> key to still place waypoints when enabled
- Requires new `tvSelection` and `tvSetSelected` script commands available on dev branch


CreepPork (approved):

neilzar (changes requested):
When the context menu is open with a unit or group selected and I right-click somewhere it will place a waypoint, I expected a new context menu to be opened or the current to be closed.

neilzar:
The previously mentioned issue is also a thing without selecting anything. Right-clicking will randomly not open the context menu and move the camera around. I think this may be an issue with incorrectly detecting the right click.

mharis001:
Can't reproduce. Moving the camera is intentional since the context menu should be opened if camera panning was started between RMB down and up events.

neilzar (approved):
I guess it is just down to the tiniest movement being detected between button-down and up.

PR #482, opened 2020-10-15
use configOf instead of typeOf config lookup

neilzar:
- Use configOf that was introduced in 2.00, it is twice as fast as using typeOf in a config lookup.


mharis001:
Should also bump `REQUIRED_VERSION`.

mharis001 (approved):

CreepPork (approved):

PR #487, opened 2020-11-16
Remove Game Master module locality transfer fix timeout

mharis001:
- title, with a very large amount of mods the module can take a long time to be unassigned when loading in initially (from testing, as much as 50 s)


CreepPork (commented):
Is there any other alternatives, by not relying on hard-coding the timeout?

mharis001:
Removing it altogether. Looking at it now, don't think it is needed.

CreepPork (approved):

neilzar (approved):

PR #494, opened 2020-12-24
Fix custom modules not being added in ZGM missions

mharis001:
- Fix custom modules not being added to modules tree in ZGM missions

### DIFF
--- a/addons/editor/functions/fnc_handleLoad.sqf
+++ b/addons/editor/functions/fnc_handleLoad.sqf
@@ -128,10 +128,19 @@ GVAR(iconsVisible) = true;
 
 [{
     // For compatibility with Zeus Game Master missions, wait until the respawn placement phase is complete
+    // and the create trees have been refreshed after curator addons are changed
     [{
         params ["_display"];
 
-        isNull _display || {entities "ModuleMPTypeGameMaster_F" isEqualTo []} || {GETMVAR(BIS_moduleMPTypeGameMaster_init,false)}
+        isNull _display
+        || {entities "ModuleMPTypeGameMaster_F" isEqualTo []}
+        || {
+            GETMVAR(BIS_moduleMPTypeGameMaster_init,false)
+            && {
+                (_display displayCtrl IDC_RSCDISPLAYCURATOR_CREATE_MODULES) tvCount [] > 1
+                || {"CuratorModeratorRights" call BIS_fnc_getParamValue < 1 && {player != BIS_curatorUnit}}
+            }
+        }
     }, {
         params ["_display"];
 
--- a/addons/editor/functions/fnc_handleLoad.sqf
+++ b/addons/editor/functions/fnc_handleLoad.sqf
@@ -137,6 +137,8 @@ GVAR(iconsVisible) = true;
 
         if (isNull _display) exitWith {};
 
+        [QGVAR(treesLoaded), _display] call CBA_fnc_localEvent;
+
         [_display] call FUNC(addGroupIcons);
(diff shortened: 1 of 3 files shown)


neilzar (approved):

Kexanone (changes requested):
Nope

<details>
<summary>Recorded Test</summary>

https://user-images.githubusercontent.com/18646707/104492843-818f1e00-55d4-11eb-93e8-710323723ef8.mp4

</details>

mharis001:
Yeah, doesn't work on dedicated since the curator addons takes time and the tree isn't refreshed in the next frame. Looking at it now, I am unsure about if there is a good way to support ZGM missions.

Kexanone:
Yeah, it's an issue that has troubled us a lot in the past.
It's far from perfect, but Achilles basically checked for the presence of ZGM and counted the entries in the module tree.
Adapting this approach to unscheduled environment might be the best approach:

````sqf
sleep 0.001;
// For Zeus Game Master missions: Wait until respawn was placed
if (!(missionNamespace getVariable ["BIS_moduleMPTypeGameMaster_init", false]) && {count allMissionObjects "ModuleMPTypeGameMaster_F" > 0}) then
{
    waitUntil {sleep 0.1; _moduleTreeCtrl tvCount [] > 1 || isNull findDisplay IDD_RSCDISPLAYCURATOR};
};
if (isNull findDisplay IDD_RSCDISPLAYCURATOR) exitWith {};

// wait for the next frame
sleep 0.001;
[Achilles_fnc_onModuleTreeLoad, []] call CBA_fnc_directCall;
````

**Comment by Kexanone** [comment]
If I remember correctly `BIS_moduleMPTypeGameMaster_init` gets set to `true` by the ZGM module when the respawn is placed, but due to race conditions, you still have to check the number of entries in the module tree.

**Comment by Kexanone** [comment]
I'm also not sure if both "delay by one frame with `sleep 0.001`" are really necessary. I might have just tested everything possible and left some unnecessary stuff around.

File: addons/editor/functions/fnc_handleLoad.sqf [hunk 1]
```diff
@@ -128,10 +128,19 @@ GVAR(iconsVisible) = true;
...
         params ["_display"];
 
-        isNull _display || {entities "ModuleMPTypeGameMaster_F" isEqualTo []} || {GETMVAR(BIS_moduleMPTypeGameMaster_init,false)}
+        isNull _display
+        || {entities "ModuleMPTypeGameMaster_F" isEqualTo []}
+        || {
+            GETMVAR(BIS_moduleMPTypeGameMaster_init,false)
+            && {
+                (_display displayCtrl IDC_RSCDISPLAYCURATOR_CREATE_MODULES) tvCount [] > 1
+                || {"CuratorModeratorRights" call BIS_fnc_getParamValue < 1 && {player != BIS_curatorUnit}}
```
Kexanone, line 141: Checking for an empty module tree (_i.e._ `tvCount` returns `0`) should already be enough for moderators without being too mission specific.

Kexanone:
I can confirm that it works now for dedicated servers too.

addons/editor/functions/fnc_handleLoad.sqf
mharis001: This line is for when moderators have no addons and can only spectate or edit existing objects. In that case, the tree is always empty.

Kexanone: Nvm, leave it as it is. My suggestion doesn't even work.

Kexanone (approved):

PR #499, opened 2021-01-08
Improve object serialization by saving vector dir and up

### LINKED ISSUE #443: Compositions not saving pitch (X) and roll (Y) of entities
**Arma 3 Version:** `1.98` (stable)
**CBA Version:** `3.15.1` (stable)
**ZEN Version:** `1.8.0` (stable)

**Mods:**
```
- CBA_A3
- ZEN
```

**Description:**
When creating a new composition which contains entities that have been rotated (e.g. a wall rolled on a 90 degree angle) once spawned from the custom composition will revert any rotation back to its default value of 0, with the exception of the yaw rotation.

**Steps to reproduce:**
1. Place an entity in the Editor. 
2. Play the mission in anywhere where you are able to use the Zeus interface.
3. Go to Groups > Empty in the Zeus interface, click on "Save new custom composition from currently selected entities." while highlighting entities that you want to save as a composition.
4. Enter name and category for the composition.
5. Select the composition under Groups > Empty > Compositions > Custom > (Category in which it was saved) > (Name of composition that was just saved).
6. Click where you want the composition to spawn.
7. Any entities that had any pitch of roll values when being turned into a composition have been reset to default.

- Dedicated, Self-Hosted Multiplayer, Singleplayer, Editor (Singleplayer), Editor (Multiplayer)

**Additional information:**
Custom compositions remember the yaw rotation of objects (Z), but not pitch (X) or roll (Y), whatever is used to remember the yaw rotation should also be added for pitch and roll.

**Screenshots:**
The wall on the left is the source entity from which a composition is saved, the right wall is that composition being spawned and its rotation values no longer being the same as the source entity, as it spawns with the default values of 0 for pitch and roll.

### LINKED ISSUE #444: Compositions do not retain Object Specific changes made to entities.
**Arma 3 Version:** `1.98` (stable)
**CBA Version:** `3.15.1` (stable)
**ZEN Version:** `1.8.0` (stable)

**Mods:**
```
- CBA_A3
- ZEN
```

**Description:**
When creating a composition in Zeus that contains an entity that has Object Specific configurations (e.g. hiding one of the draws on a drawer prop) will not carry through when that object is saved and spawned as a custom composition.

**Steps to reproduce:**
Place an entity in the Editor which has Object Specific configurations (e.g. Rugged Portable Desk).
Open the attributes of the entity and make changes under Object Specific and press OK.
Play the mission in anywhere where you are able to use the Zeus interface.
Go to Groups > Empty in the Zeus interface, click on "Save new custom composition from currently selected entities." while highlighting entity that you want to save as a composition.
Enter name and category for the composition.
Select the composition under Groups > Empty > Compositions > Custom > (Category in which it was saved) > (Name of composition that was just saved).
Click where you want the composition to spawn.
Any entity that had any changes made under Object Specific will reset to default values when spawned as a composition).

- Dedicated, Self-Hosted Multiplayer, Singleplayer, Editor (Singleplayer), Editor (Multiplayer)

**Additional information:**
Any entity that has had changes made to it under Object Specific in it's attributes while in Editor need some way of carrying that information over when saved and spawned as a composition through the Zeus interface.

**Screenshots:**
The Military Cargo Platform allows you to configure the height of each of the four legs, adjust the rotation of each of the four side panels or hide any of the four panels completely, however once the source object (Left) has been saved and spawned as a composition, it will revert everything back to default.

### LINKED ISSUE #455: Unit identity not getting set on dedicated servers (applied too early?)
**Description:**
When you spawn a custom composition or deep paste units on a dedicated server, where it usually takes 1-2 seconds for the units to load properly and their custom gear to be applied, their custom identity (face, insignia, etc) will not be loaded.

When doing the same on a local game, it works fine as the units are (unlike on dedicated servers) loaded immediately. Thus I think ZEN is applying the custom identity too early and needs to wait until the saved gear etc for the spawned unit has finished loading.

**Steps to reproduce:**
- Save a custom composition with units using set faces/insignias/etc
- Place the composition on a local game -> See that the identity did apply
- Place the composition on a dedicated server -> See that the identity didn't apply
- Repeat with a unit using deep copy and paste

- Dedicated

Kexanone:
- [x]  Fix #443


neilzar (approved):
Maybe worth adding a deprecation warning

Kexanone:
> Maybe worth adding a deprecation warning

No, need. We just accept both formats. It would be rather annoying if you get a deprecation warning for all your compositions.

mharis001 (approved):

PR #501, opened 2021-01-09
Add Saved selection for modules and Transfer Ownership Module

ampersand38:
- Add global variable `zen_editor_savedSelection` which tracks the previous-to-last content of `curatorSelected`, used for selecting entities then dropping a module affecting them.
- Add Transfer Ownership module using saved selection
- https://youtu.be/4L-czAJIIcQ

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -314,8 +314,36 @@ if (isServer) then {
     }] call CBA_fnc_addEventHandler;
 
     [QGVAR(enableSimulationGlobal), {
+        diag_log _this;
         params ["_object", "_enable"];
         _object enableSimulationGlobal _enable;
+    }] call CBA_fnc_addEventHandler;
+
+    [QGVAR(transferOwnership), {
+        diag_log QGVAR(transferOwnership);
+        diag_log _this;
+        params ["_entities", "_target"];
+        if (!(_entities isEqualType [])) then {
+            _entities = [_entities];
+        };
+        private _clientID = 0;
+        if (_target isEqualType 0) then {
+            _clientID = _target;
+        };
+        if (_target isEqualType objNull) then {
+            _clientID = owner _target;
+        };
+        {
+            if (_x isEqualType grpNull) then {
+                _x setGroupOwner _clientID;
+            } else {
+                if (group _x == grpNull) then {
+                    group _x setGroupOwner _clientID;
+                } else {
+                    _x setOwner _clientID;
+                };
+            };
+        } forEach _entities;
     }] call CBA_fnc_addEventHandler;
 
     [QGVAR(setFriend), {
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -314,14 +314,11 @@ if (isServer) then {
     }] call CBA_fnc_addEventHandler;
 
     [QGVAR(enableSimulationGlobal), {
-        diag_log _this;
         params ["_object", "_enable"];
         _object enableSimulationGlobal _enable;
     }] call CBA_fnc_addEventHandler;
 
     [QGVAR(transferOwnership), {
-        diag_log QGVAR(transferOwnership);
-        diag_log _this;
         params ["_entities", "_target"];
         if (!(_entities isEqualType [])) then {
             _entities = [_entities];
--- a/addons/editor/functions/fnc_handleSelectionChanged.sqf
+++ b/addons/editor/functions/fnc_handleSelectionChanged.sqf
@@ -1,0 +1,23 @@
+#include "script_component.hpp"
+/*
+ * Author: Ampersand
+ * Saves curator selected entities for use by modules.
+ *
+ * Arguments:
+ * 0: Curator <OBJECT>
+ * 0: Entity: group, marker, object, or waypoint <GROUP, STRING, OBJECT, ARRAY>
+ *
+ * Return Value:
+ * Handled <BOOL>
+ *
+ * Example:
+ * [CONTROL] call zen_editor_fnc_handleSelectionChanged
+ *
+ * Public: No
+ */
+
+params ["_curator", "_entity"];
+GVAR(savedSelection) = GVAR(lastSelection);
+GVAR(lastSelection) = curatorSelected;
+
+false
--- a/addons/modules/functions/fnc_moduleTransferOwnership.sqf
+++ b/addons/modules/functions/fnc_moduleTransferOwnership.sqf
@@ -1,0 +1,72 @@
+#include "script_component.hpp"
+/*
+ * Author: Ampersand
+ * Zeus module function to transfer ownership of objects.
+ *
+ * Arguments:
+ * 0: Logic <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [LOGIC] call zen_modules_fnc_moduleTransferOwnership
+ *
+ * Public: No
+ */
+
+if (!isMultiplayer) exitWith {
+ [LSTRING(OnlyMultiplayer)] call EFUNC(common,showMessage);
+};
+
+params ["_logic"];
+
+private _unit = attachedTo _logic;
+deleteVehicle _logic;
+
+private _clientTypes = allPlayers;
+private _clientNames = allPlayers apply {name _x};
+if (!isServer) then {
+    _clientTypes = [2] + _clientTypes;
+    _clientNames = ["Server"] + _clientNames;
+};
+
+private _entities = [];
+private _mehID = -1;
+if (isNull _unit) then {
+    _mehID = ["\a3\ui_f\data\Map\VehicleIcons\iconVirtual_ca.paa"] call EFUNC(editor,drawSavedSelectionIcons);
+    EGVAR(editor,savedSelection) params ["_objects", "_groups"];
+    {
+        if (isNull group _x) then {
+            _entities pushBack _x;
+        } else {
+            _groups pushBackUnique group _x;
+        };
+    } forEach _objects;
+    _entities append _groups;
+} else {
+    _entities = [[group _unit, _unit] select (isNull group _unit)];
+};
+
+[LSTRING(ModuleTransferOwnership), [
+    [
+        "COMBO",
+        ELSTRING(common,Target),
+        [_clientTypes, _clientNames, 0]
+    ]
+], {
+    params ["_values", "_args"];
+    _args params ["_entities", "_mehID"];
+    if (_mehID > 0) then {
+        removeMissionEventHandler ["Draw3D", _mehID];
+    };
+
+    _values params ["_target"];
+    [QEGVAR(common,transferOwnership), [_entities, _target]] call CBA_fnc_serverEvent;
+}, {
+    params ["", "_args"];
+    _args params ["", "_mehID"];
+    if (_mehID > 0) then {
+        removeMissionEventHandler ["Draw3D", _mehID];
+    };
+}, [_entities, _mehID]] call EFUNC(dialog,create);
--- a/addons/modules/functions/fnc_moduleTransferOwnership.sqf
+++ b/addons/modules/functions/fnc_moduleTransferOwnership.sqf
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 /*
  * Author: Ampersand
- * Zeus module function to transfer ownership of objects.
+ * Zeus module function to transfer ownership of objects and groups.
  *
  * Arguments:
  * 0: Logic <OBJECT>
@@ -24,17 +24,9 @@ params ["_logic"];
 private _unit = attachedTo _logic;
 deleteVehicle _logic;
 
-private _clientTypes = allPlayers;
-private _clientNames = allPlayers apply {name _x};
-if (!isServer) then {
-    _clientTypes = [0] + _clientTypes;
-    _clientNames = ["str_a3_om_common_definitions.incphone_44"] + _clientNames;
-};
-
 private _entities = [];
-private _mehID = -1;
 if (isNull _unit) then {
-    _mehID = ["\a3\ui_f\data\Map\VehicleIcons\iconVirtual_ca.paa"] call EFUNC(editor,drawSavedSelectionIcons);
+    // No attached unit, get saved selection
     EGVAR(editor,savedSelection) params ["_objects", "_groups"];
     {
         if (isNull group _x) then {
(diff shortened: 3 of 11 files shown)


oOKexOo:
Can you enable edits from maintainers?
Wanted to add the revised selection preview.

jonpas:
Is that naked unit bug fix basically same as in ACEX or did you improve it in any way?

oOKexOo:
> Is that naked unit bug fix basically same as in ACEX Headless or did you improve it in any way?
Oh, I forgot about the ACEX one. I basically improved the Achilles' version of the fix by switching to unscheduled and reducing network traffic by not broadcasting the loadouts to all clients.

> Side-note: It would be nice to have ACEX Headless compatibility when transfering to client (automatically put into headless blacklist for automatic transfers). Or an API event to be able to do it in a mission framework for example.
Yeah, we still have to look into this.

jonpas:
> I basically improved the Achilles' version of the fix by switching to unscheduled and reducing network traffic by not broadcasting the loadouts to all clients.

ACEX Headless does transmit loadout to everyone as well, and that is for sure an improvement in network traffic. But I don't see how it can be enough of an improvement to "fix" the issue, not while there are a million way more intensive things being broadcasted.

This seems to be a possible better fix (which I haven't implemented in ACEX yet): https://github.com/acemod/ACEX/issues/180#issuecomment-662517165 (also ref. linked BI ticket).

oOKexOo:
> ACEX Headless does transmit loadout to everyone as well, and that is for sure an improvement in network traffic. But I don't see how it can be enough of an improvement to "fix" the issue, not while there are a million way more intensive things being broadcasted.

I agree with that, but the more important improvement of the new implementation is that the loadouts are guaranteed to be defined when you restore them. Achilles did not check and wait for the loadouts broadcasted via public `setVariable` to be defined on the new owner.

jonpas:
Ah I see what you mean. That's actually a pretty good point.

oOKexOo:
Actually, seeing now the ACEX implementation. The `local` EH combined with the appropriate checks via `CBA_fnc_waitUntilAndExecute` could be the best approach to tackle this problem in a general way. At this point, why not provide a common fix for all of us with CBA?

jonpas:
> At this point, why not provide a common fix for all of us with CBA?

Sure, I think that would be in scope. Want to open a PR?

oOKexOo:
> Sure, I think that would be in scope. Want to open a PR?

Yeah, I'll work on one when I got time.
Would fit nicely into the new `characters` addon if you merge CBATeam/CBA_A3#1344 ;)

jonpas:
Awesome! :+1: 

Yeah, it's tagged for 3.16. But things are slow in this age of Arma 3.

Kexanone:
Reverted the fix in favor of CBATeam/CBA_A3#1406

ampersand38:
image

addons/modules/functions/fnc_moduleTransferOwnership.sqf
```diff
@@ -81,6 +91,16 @@ _clientNames = ["str_a3_om_common_definitions.incphone_44"] + _clientNames;
             _player
         };
     };
+
+    // set headless client script flags
+    if (_HCState < 2) then {
+        if (acex_headless) then {
```
jonpas, line 97: This won't be `false` if ACEX is not loaded. You probably want to check the config class.

ampersand38: 30ad1c6

addons/modules/stringtable.xml
```diff
@@ -592,15 +592,28 @@
...
+            <Russian>Передать владение</Russian>
+            <German>Besitztum verschieben</German>
+            <Japanese>所有権を移行</Japanese>
+            <Chinesesimp>转让所有权</Chinesesimp>
+            <Chinese>轉讓所有權</Chinese>
         </Key>
         <Key ID="STR_ZEN_Modules_ModuleTransferOwnership_Client">
             <English>Client</English>
+            <French>Client</French>
+            <German>Klient<German>
```
neilzar, line 606: Missing `/` in close tag

addons/modules/stringtable.xml
```diff
@@ -592,15 +592,28 @@
...
         </Key>
         <Key ID="STR_ZEN_Modules_ModuleTransferOwnership_Server">
             <English>Server</English>
+            <French>Serveur</French>
+            <German>Server<German>
```
neilzar, line 611: Same as above

addons/modules/stringtable.xml
```diff
@@ -592,15 +592,28 @@
...
         </Key>
         <Key ID="STR_ZEN_Modules_ModuleTransferOwnership_HCScripts">
             <English>HC Scripts</English>
+            <French>Script pour HC</French>
+            <German>HC Skript<German>
```
neilzar, line 616: Same as above

neilzar (changes requested):

neilzar (approved):

addons/common/XEH_postInit.sqf
```diff
@@ -318,6 +318,31 @@ if (isServer) then {
...
+            _clientID = _target;
+        };
+        if (_target isEqualType objNull) then {
+            _clientID = owner _target;
+        };
+        {
+            if (_x isEqualType grpNull) then {
+                _x setGroupOwner _clientID;
+            } else {
+                if (group _x == grpNull) then {
```
Kexanone, line 337: I think you messed up the condition here. Group will be null for objects.

Kexanone:
For now, just focus on the transfer ownership module, which seems to work fine besides the naked unit issue that will be handled by CBA in future.
We should make a separate PR for generalizing and applying the selection feature to other modules.

addons/editor/functions/fnc_toggleSelectionPreview.sqf
```diff
@@ -0,0 +1,52 @@
...
+                        } forEach (GVAR(lastSelection) select 0);
+                    }];
+                };
+            } else {
+                if !(isNil QGVAR(drawSelectioIcons)) then {
+                    removeMissionEventHandler ["Draw3D", GVAR(drawSelectioIcons)];
+                    GVAR(drawSelectioIcons) = nil
+                };
+            };
+        }, 0] call CBA_fnc_addPerFrameHandler;
```
mharis001, line 41: Instead of a PFH to add/remove Draw3D handler, use a Draw3D handler that doesn't draw if an entity is hovered.

addons/modules/functions/fnc_moduleTransferOwnership.sqf
```diff
@@ -0,0 +1,113 @@
...
+    [QEGVAR(common,transferOwnership), [_entities, _targetID]] call CBA_fnc_serverEvent;
+}, {
+    params ["", "_args"];
+    _args params ["", "_mehID"];
+
+    // Stop drawing icons
+    if (_mehID > 0) then {
+        removeMissionEventHandler ["Draw3D", _mehID];
+    };
+}, [_entities, _mehID]] call EFUNC(dialog,create);
```
mharis001, line 113: `_mehID` is undefined.

addons/modules/functions/fnc_moduleTransferOwnership.sqf
```diff
@@ -0,0 +1,113 @@
...
+            "str_a3_cfgvehicles_module_f_moduledescription_curator_f_1",
+            LSTRING(ModuleTransferOwnership_Client)
+        ]],
+        true
+    ],
+    [
+        "COMBO",
+        LSTRING(ModuleTransferOwnership_Client),
+        [_clientTypes, _clientNames, 0]
+    ],
```
mharis001, line 63: I think these two options control the same thing and should be combined into a single list. The list would contain the server (if not hosted MP) and a list of clients. The curator and server can be highlighted.

addons/modules/functions/fnc_moduleTransferOwnership.sqf
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+params ["_logic"];
+
+if (!isMultiplayer) exitWith {
+    [LSTRING(OnlyMultiplayer)] call EFUNC(common,showMessage);
+    deleteVehicle _logic;
+};
+
+deleteVehicle _logic;
```
mharis001, line 25: It's deleted in both cases. Do that before the `if`.

addons/modules/functions/fnc_moduleTransferOwnership.sqf
```diff
@@ -0,0 +1,113 @@
...
+{
+    if (isNull group _x) then {
+        _entities pushBack _x;
+    } else {
+        _groups pushBackUnique group _x;
+    };
+} forEach _objects;
+_entities append _groups;
+
+if (_entities findIf {units _x findIf {isPlayer _x} > -1} > -1) exitWith {
```
mharis001, line 38: `!= -1`.

addons/editor/functions/fnc_handleSelectionChanged.sqf
```diff
@@ -0,0 +1,23 @@
+#include "script_component.hpp"
+/*
+ * Author: Ampersand
+ * Saves curator selected entities for use by modules.
+ *
+ * Arguments:
+ * 0: Curator <OBJECT>
+ * 0: Entity: group, marker, object, or waypoint <GROUP, STRING, OBJECT, ARRAY>
```
mharis001, line 8: `OBJECT|GROUP|ARRAY|STRING`.

addons/editor/functions/fnc_handleSelectionChanged.sqf
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Return Value:
+ * Handled <BOOL>
```
mharis001, line 11: Can this event handler be overridden?

mharis001 (changes requested):

addons/editor/functions/fnc_handleSelectionChanged.sqf
Kexanone: No, we don't store the index of the event anywhere, but we did the same for all other curator event handlers. We could consider bridging them with CBA event handlers, but I would make a separate PR for such a change.

ampersand38:
https://youtu.be/sAHKlZA-Vz4

addons/modules/functions/fnc_moduleTransferOwnership.sqf
```diff
@@ -0,0 +1,106 @@
...
+private _entities = [];
+(call EFUNC(editor,getSelection)) params ["_objects", "_groups"];
+{
+    if (isNull group _x) then {
+        _entities pushBack _x;
+    } else {
+        _groups pushBackUnique group _x;
+    };
+} forEach _objects;
+_entities append _groups;
```
freghar, line 34: Won't this duplicate some groups (added once via `_object` iteration, appended via `_groups`)?

neilzar (approved):

addons/modules/functions/fnc_moduleTransferOwnership.sqf
```diff
@@ -0,0 +1,106 @@
...
+
+params ["_logic"];
+deleteVehicle _logic;
+
+if (!isMultiplayer) exitWith {
+    [LSTRING(OnlyMultiplayer)] call EFUNC(common,showMessage);
+};
+
+private _entities = [];
+(call EFUNC(editor,getSelection)) params ["_objects", "_groups"];
```
CreepPork, line 26:
```suggestion
call EFUNC(editor,getSelection) params ["_objects", "_groups"];
```

addons/modules/functions/fnc_moduleTransferOwnership.sqf
```diff
@@ -0,0 +1,106 @@
...
+// Set default target to curator, server, or HC depending on current locality
+private _defaultTarget = if (local (_entities select 0)) then {
+    if (_HCs isEqualTo []) then {
+        0
+    } else {
+        2
+    };
+} else {
+    1
+};
```
CreepPork, line 73: Use `select` syntax here.

addons/modules/functions/fnc_moduleTransferOwnership.sqf
```diff
@@ -0,0 +1,106 @@
...
+[LSTRING(ModuleTransferOwnership), [
+    [
+        "COMBO",
+        ELSTRING(common,Target),
+        [_targets, _targetNames, _defaultTarget],
+        true
+    ],
+    [
+        "TOOLBOX",
+        LSTRING(ModuleTransferOwnership_HCScripts),
```
CreepPork, line 84: Tooltip, at least, please? What is a HC script?

addons/modules/functions/fnc_moduleTransferOwnership.sqf
ampersand38: 7e3f1d1

addons/modules/functions/fnc_moduleTransferOwnership.sqf
ampersand38: It does auto-select based on the current locality:
- If local, assume target will be HC and balancing will be enabled.
- If remote, assume target will be Curator and balancing will be disabled.
However I wanted to let user pick for edge cases.
a1696ce

CreepPork (approved):

addons/common/XEH_postInit.sqf
```diff
@@ -318,6 +318,31 @@ if (isServer) then {
...
+                    _x setOwner _clientID;
+                } else {
+                    group _x setGroupOwner _clientID;
+                };
+            };
+        } forEach _entities;
+    }] call CBA_fnc_addEventHandler;
```
mharis001, line 344: I think this is big enough to warrant moving into a function/separate file.

addons/modules/functions/fnc_moduleTransferOwnership.sqf
```diff
@@ -0,0 +1,101 @@
...
+        true
+    ],
+    [
+        "TOOLBOX",
+        [LSTRING(ModuleTransferOwnership_HCScripts), LSTRING(ModuleTransferOwnership_HCScripts_Description)],
+        [parseNumber (_defaultTarget == 2), 1, 2, [
+            ELSTRING(common,Disabled),
+            ELSTRING(common,Enabled)
+        ]],
+        true
```
mharis001, line 84: Use `TOOLBOX:ENABLED` here.

addons/editor/functions/fnc_toggleSelectionPreview.sqf
```diff
@@ -0,0 +1,38 @@
...
+params [["_turnOn", true, [true]]];
+
+if (_turnOn) then {
+    if (isNil QGVAR(selectionIconHandler)) then {
+        GVAR(selectionIconHandler) = addMissionEventHandler ["Draw3D", {
+            if (curatorMouseOver isEqualTo [""]) then {
+                {
+                    drawIcon3D [
+                        GVAR(lastModuleIcon),
+                        GVAR(colour), getPosVisual _x, 1, 1, 0
```
mharis001, line 27: `getPosVisual` returns AGLS, `drawIcon3D` expects AGL.

mharis001 (changes requested):
I would like to see this feature moved to its own component so it is easier to work on in the future. `editor` component already has a lot of responsibilities and it will be difficult to see everything involved in making this work.

PR #387, opened 2020-05-21
Extended Vehicle Customization

oOKexOo:
**When merged this pull request will:**
- Provide `zen_garage_fnc_defineCustomTexture` as API function for defining custom textures for the garage.
- Unlock hidden vanilla textures for the garage.
- Unlock additional animation sources for the garage (_e.g._ vehicle doors).
- Add support for the extended customization to deep copy/paste and SQF exporter.

PR #504, opened 2021-01-12
Add Faction Filter

### LINKED ISSUE #142: Feature Request: Module & Faction Curator Restrictions
[Achilles](https://steamcommunity.com/sharedfiles/filedetails/?id=723217262) has a CBA setting which allows you to restrict the modules and factions that a curator has access to. This is really handy for streamlining the zues experience by only including relevant factions. This also prevents less experienced curators from placing wrong units. Same applies, in the modules that a curator would have access to.

It would be great to have in Zeus Enhanced and would allow my unit to finally move onto Enhanced fully.

ReferenceImage

mharis001:
- Add `faction_filter` component
    - Adds settings to control which factions (by name) are available in the unit and group trees
    - Close #142 
- Fix invisible side buttons on initial load with empty trees
    - Simplify code for `modeChanged` and `sideChanged` events

Works well when tested with popular asset mods. Deals with the following object categorization trickiness (for Zeus):
- `side` of `CfgFactionClasses` doesn't determine under which side an object is sorted.
- `editorCategory` can also be used to categorize an object (although generally not used for characters and vehicles).
- group compositions can have a different faction display name than objects from the same faction.

 As a result, we don't use `CfgFactionClasses` directly and instead scan `CfgPatches` `units` arrays and `CfgGroups` to find factions.

### DIFF
--- a/addons/editor/functions/fnc_fixSideButtons.sqf
+++ b/addons/editor/functions/fnc_fixSideButtons.sqf
@@ -5,30 +5,30 @@
  *
  * Arguments:
  * 0: Display <DISPLAY>
- * 1: Mode <NUMBER>
  *
  * Return Value:
  * None
  *
  * Example:
- * [DISPLAY, 0] call zen_editor_fnc_fixSideButtons
+ * [DISPLAY] call zen_editor_fnc_fixSideButtons
  *
  * Public: No
  */
 
-[{
-    params ["_display", "_mode"];
+params ["_display"];
 
-    // Get side buttons to show based on mode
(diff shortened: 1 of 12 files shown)


Kexanone:
Seems to work well in "vanilla", but we might want to check some more mods.
Going back to my other question: https://github.com/zen-mod/ZEN/pull/387#issuecomment-759618361

neilzar (approved):
Confirmed to work with all RHS mods and 3CB factions, so I'd say all mods should work.

Kexanone (approved):

PR #511, opened 2021-01-13
Several CI changes

neilzar:
- Update CI validation scripts to latest versions
- Fix notify to only trigger once

### DIFF
--- a/tools/stringtable_validator.py
+++ b/tools/stringtable_validator.py
@@ -24,7 +24,7 @@ def check_stringtable(filepath):
     try:
         tree = ET.parse(filepath)
     except Exception as e:
-        print("  ERROR: Failed to parse file. {}".format(e))
+        print("  ERROR: Failed to parse file.")
         return 1
 
     errors = 0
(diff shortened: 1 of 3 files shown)


tools/stringtable_validator.py
```diff
@@ -24,7 +24,7 @@ def check_stringtable(filepath):
     try:
         tree = ET.parse(filepath)
     except Exception as e:
-        print("  ERROR: Failed to parse file. {}".format(e))
+        print("  ERROR: Failed to parse file.")
```
mharis001, line 27: We want to keep the excepting printing.

tools/stringtable_validator.py
```diff
@@ -162,13 +162,11 @@ def main():
...
     for filepath in stringtable_files:
-        print("\nChecking {}:".format(os.path.relpath(filepath, root_dir)))
+        print("Checking {}:".format(os.path.relpath(filepath, root_dir)))
 
         errors = check_stringtable(filepath)
 
-        if errors == 0:
-            print("No errors found.")
-        else:
+        if errors != 0:
```
mharis001, line 169: I'd prefer to keep the more nicely formatted output. We don't have too many components so it works fine.

mharis001 (approved):

Kexanone (approved):

PR #512, opened 2021-01-14
Add HEMTT to build.py

neilzar:
- Use HEMTT when running build.py, if it is installed.

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -14,7 +14,7 @@ def tryHemttBuild(projectpath):
     if os.path.isfile(hemttExe):
         os.chdir(projectpath)
         ret = subprocess.call([hemttExe, "pack"], stderr=subprocess.STDOUT)
-        print("Using hemtt: {}".format(ret));
+        print("Using HEMTT: {}".format(ret));
         return True
     else:
         print("HEMTT is not available");
--- a/tools/build.py
+++ b/tools/build.py
@@ -9,6 +9,17 @@ MAINPREFIX = "x"
 PREFIX = "zen_"
 ##########################
 
+def tryHemttBuild(projectpath):
+    hemttExe = os.path.join(projectpath, "hemtt.exe")
+    if os.path.isfile(hemttExe):
+        os.chdir(projectpath)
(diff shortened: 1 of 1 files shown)


tools/build.py
```diff
@@ -9,6 +9,17 @@
 PREFIX = "zen_"
 ##########################
 
+def tryHemttBuild(projectpath):
+    hemttExe = os.path.join(projectpath, "hemtt.exe")
```
CreepPork, line 13: What if I would want to build it on Linux too?

tools/build.py
```diff
@@ -9,6 +9,17 @@
...
+    if os.path.isfile(hemttExe):
+        os.chdir(projectpath)
+        ret = subprocess.call([hemttExe, "pack"], stderr=subprocess.STDOUT)
+        print("Using hemtt: {}".format(ret));
+        return True
+    else:
+        print("hemtt not installed");
```
CreepPork, line 20:
```suggestion
        print("HEMTT is not available");
```

tools/build.py
```diff
@@ -9,6 +9,17 @@
 PREFIX = "zen_"
 ##########################
 
+def tryHemttBuild(projectpath):
+    hemttExe = os.path.join(projectpath, "hemtt.exe")
+    if os.path.isfile(hemttExe):
+        os.chdir(projectpath)
+        ret = subprocess.call([hemttExe, "pack"], stderr=subprocess.STDOUT)
+        print("Using hemtt: {}".format(ret));
```
CreepPork, line 17:
```suggestion
        print("Using HEMTT: {}".format(ret));
```

tools/build.py
neilzar: I took this code from the CBA repo, I don't have enough experience with Linux and its file system, or with python, to know how to make it work.

jonpas: You shouldn't use `build.py`, but use `hemtt` binary directly.

CreepPork: Yep, that's true. I'd guess I'm just overthinking it too much.

CreepPork (approved):

Kexanone (approved):

mharis001 (approved):

PR #516, opened 2021-01-18
Fix RC breaking when entering ACE arsenal

### LINKED ISSUE #489: Editing a remote controlled units inventory breaks RC
**Arma 3 Version**: 2.00 (stable)
**CBA Version**: 3.15.2.201119 (stable)
**ZEN Version**: 1.9.0 (stable)

**Mods:**
- CBA_A3
- ZEN
- ZEUS CACHE
- RHS USAF, AFRF, GREF
- TFAR BETA
- ACE
- VCOM 3.4.0
- PROJECT OPFOR
- ENHANCED MOVEMENT
- SPLENDID SMOKE

**Description:**

> When myself and my friend were co zeusing a mission for our group we encountered a pretty severe bug. 
> when i remote controlled a unit and asked my friend to give me a weapon (I think rangefinder in this case) 
> I could tell when he accessed my unit as my guy froze. Anyway.. when he said ok done my guy was frozen and could not control him anymore. 
> 
> After exiting the remote control of that specific unit, any unit I tried to RC would not fully get taken over. Instead they proceed to have full control with movement and firing. 
> 
> I then decided to leave the server and re join. When I re joined I lost all Zeus capabilities. 
> I would also like to add that as soon as we promoted eachother zeus as it happened to both of us at different times of the mission. ZEUS UI LOCKUP happened from then onward. 
> 
> Its linked to getting zeus while in game I think.

**Add the steps needed to reproduce the issue.**

> Get 1 to RC a opfor unit and get the other zeus to edit inventory and give him a rangefinder or something and then back out of it releasing him.

> On a Dedicated Server (Stable exe)

**Additional information:** 

> - Neither of us was on Profiling branch

- 

**Provide any additional information that will help us solve this issue.
RPT log file:** 

> Unfortunately I was unable to locate the rpt since we have done multiple missions since. 

**### Original post below.**

When 2 people are zeusing 

Get 1 to RC a opfor unit and get the other zeus to edit inventory and give him a launcher or something and then back out of it releasing him. 

His unit will lockup and whenever he tries to RC a unit afterwards he cannot control them. He has to leave the server and re join. Even after doing so his zeus then does not work at all.

neilzar:
- Fixes #489

### DIFF
--- a/addons/common/functions/fnc_isRemoteControlled.sqf
+++ b/addons/common/functions/fnc_isRemoteControlled.sqf
@@ -1,0 +1,21 @@
+#include "script_component.hpp"
+/*
+ * Author: NeilZar
+ *
+ * Arguments:
+ * 0: Unit <OBJECT>
+ *
+ * Return Value:
+ * Unit is Remote Controlled <BOOL>
+ *
+ * Example:
+ * [cursorObject] call zen_common_fnc_isRemoteControlled
+ *
+ * Public: Yes
+ */
+
+params ["_unit"];
+
+if !(_unit isKindOf "CAManBase") exitWith {false};
+
+allCurators findIf {_x getVariable "bis_fnc_moduleremotecontrol_unit" isEqualTo _unit} != -1
--- a/addons/common/functions/fnc_isRemoteControlled.sqf
+++ b/addons/common/functions/fnc_isRemoteControlled.sqf
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 /*
  * Author: NeilZar
- * Check if the given unit is currently remote controlled.
+ * Checks if the given unit is currently remote controlled.
  *
  * Arguments:
  * 0: Unit <OBJECT>
@@ -17,6 +17,6 @@
 
 params ["_unit"];
 
-_curators = allCurators apply {getAssignedCuratorUnit _x} select {!(isNull _x)};
+private _curators = allCurators apply {getAssignedCuratorUnit _x} select {!(isNull _x)};
 
 _unit getVariable ["bis_fnc_moduleremotecontrol_owner", objNull] in _curators
--- a/addons/common/functions/fnc_isRemoteControlled.sqf
+++ b/addons/common/functions/fnc_isRemoteControlled.sqf
@@ -1,6 +1,7 @@
 #include "script_component.hpp"
 /*
  * Author: NeilZar
+ * Check whether the given unit is currently remote controlled.
  *
  * Arguments:
  * 0: Unit <OBJECT>
@@ -9,7 +10,7 @@
  * Unit is Remote Controlled <BOOL>
  *
  * Example:
- * [cursorObject] call zen_common_fnc_isRemoteControlled
+ * [_unit] call zen_common_fnc_isRemoteControlled
  *
  * Public: Yes
  */
@@ -18,4 +19,4 @@ params ["_unit"];
 
 if !(_unit isKindOf "CAManBase") exitWith {false};
 
-allCurators findIf {_x getVariable "bis_fnc_moduleremotecontrol_unit" isEqualTo _unit} != -1
+_unit getVariable ["bis_fnc_moduleremotecontrol_owner", objNull] in allCurators
--- a/addons/common/functions/fnc_isRemoteControlled.sqf
+++ b/addons/common/functions/fnc_isRemoteControlled.sqf
@@ -17,6 +17,4 @@
 
 params ["_unit"];
 
-private _curators = allCurators apply {getAssignedCuratorUnit _x} select {!(isNull _x)};
-
-_unit getVariable ["bis_fnc_moduleremotecontrol_owner", objNull] in _curators
+!isNull (_unit getVariable ["bis_fnc_moduleremotecontrol_owner", objNull])
--- a/addons/common/functions/fnc_isRemoteControlled.sqf
+++ b/addons/common/functions/fnc_isRemoteControlled.sqf
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 /*
  * Author: NeilZar
- * Check whether the given unit is currently remote controlled.
+ * Check if the given unit is currently remote controlled.
  *
  * Arguments:
  * 0: Unit <OBJECT>
@@ -17,6 +17,6 @@
 
 params ["_unit"];
 
-if !(_unit isKindOf "CAManBase") exitWith {false};
+_curators = allCurators apply {getAssignedCuratorUnit _x} select {!(isNull _x)};
 
-_unit getVariable ["bis_fnc_moduleremotecontrol_owner", objNull] in allCurators
+_unit getVariable ["bis_fnc_moduleremotecontrol_owner", objNull] in _curators
--- a/addons/common/functions/fnc_openArsenal.sqf
+++ b/addons/common/functions/fnc_openArsenal.sqf
@@ -17,7 +17,7 @@
 
 params ["_unit"];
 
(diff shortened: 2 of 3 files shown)


neilzar:
This is a possible solution, could also go with not opening the ACE arsenal and show a warning. I thought this was a bit more elegant solution.

addons/common/functions/fnc_openArsenal.sqf
```diff
@@ -17,7 +17,7 @@
 
 params ["_unit"];
 
-if (GVAR(preferredArsenal) == 1 && {isClass (configFile >> "CfgPatches" >> "ace_arsenal")}) then {
+if (GVAR(preferredArsenal) == 1 && {isClass (configFile >> "CfgPatches" >> "ace_arsenal")} && {!([_entity] call zen_common_fnc_isRemoteControlled)}) then {
```
mharis001, line 20: `FUNC` macro.

addons/common/functions/fnc_isRemoteControlled.sqf
```diff
@@ -0,0 +1,21 @@
...
+ * [cursorObject] call zen_common_fnc_isRemoteControlled
+ *
+ * Public: Yes
+ */
+
+params ["_unit"];
+
+if !(_unit isKindOf "CAManBase") exitWith {false};
+
+allCurators findIf {_x getVariable "bis_fnc_moduleremotecontrol_unit" isEqualTo _unit} != -1
```
mharis001, line 21: This variable is set on `missionNamespace`. I think you want to use `bis_fnc_moduleRemoteControl_owner` instead.

addons/common/functions/fnc_isRemoteControlled.sqf
```diff
@@ -0,0 +1,21 @@
+#include "script_component.hpp"
+/*
+ * Author: NeilZar
+ *
```
mharis001, line 4: Missing description.

addons/common/functions/fnc_isRemoteControlled.sqf
```diff
@@ -0,0 +1,21 @@
...
+ * Arguments:
+ * 0: Unit <OBJECT>
+ *
+ * Return Value:
+ * Unit is Remote Controlled <BOOL>
+ *
+ * Example:
+ * [cursorObject] call zen_common_fnc_isRemoteControlled
```
mharis001, line 12: I'd prefer `_unit`.

mharis001 (changes requested):

addons/common/functions/fnc_isRemoteControlled.sqf
```diff
@@ -0,0 +1,22 @@
+#include "script_component.hpp"
```
mharis001, line 1: Function is missing `PREP` entry.

addons/common/functions/fnc_isRemoteControlled.sqf
```diff
@@ -0,0 +1,22 @@
...
+/*
+ * Author: NeilZar
+ * Check whether the given unit is currently remote controlled.
```
mharis001, line 4:
```suggestion
 * Checks if the given unit is currently remote controlled.
```

addons/common/functions/fnc_isRemoteControlled.sqf
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Example:
+ * [_unit] call zen_common_fnc_isRemoteControlled
+ *
+ * Public: Yes
+ */
+
+params ["_unit"];
+
+if !(_unit isKindOf "CAManBase") exitWith {false};
```
mharis001, line 20: Is this check necessary? If it is, combine with the one below using `&&`.

addons/common/functions/fnc_isRemoteControlled.sqf
```diff
@@ -0,0 +1,22 @@
...
+
+_unit getVariable ["bis_fnc_moduleremotecontrol_owner", objNull] in allCurators
```
mharis001, line 22: Variable is set as the controlling player unit, not the controlling curator logic.

addons/common/functions/fnc_openArsenal.sqf
```diff
@@ -17,7 +17,7 @@
 
 params ["_unit"];
 
-if (GVAR(preferredArsenal) == 1 && {isClass (configFile >> "CfgPatches" >> "ace_arsenal")}) then {
+if (GVAR(preferredArsenal) == 1 && {isClass (configFile >> "CfgPatches" >> "ace_arsenal")} && {!([_entity] call FUNC(isRemoteControlled))}) then {
```
mharis001, line 20:
```suggestion
if (GVAR(preferredArsenal) == 1 && {isClass (configFile >> "CfgPatches" >> "ace_arsenal")} && {!(_entity call FUNC(isRemoteControlled))}) then {
```

mharis001 (changes requested):

addons/common/functions/fnc_isRemoteControlled.sqf
neilzar: The benefit of coding at 4AM, you are bound to forget things. 😅

addons/common/functions/fnc_isRemoteControlled.sqf
```diff
@@ -0,0 +1,22 @@
+#include "script_component.hpp"
+/*
+ * Author: NeilZar
+ * Check if the given unit is currently remote controlled.
```
mharis001, line 4: "Checks".

addons/common/functions/fnc_isRemoteControlled.sqf
```diff
@@ -0,0 +1,22 @@
...
+ * [_unit] call zen_common_fnc_isRemoteControlled
+ *
+ * Public: Yes
+ */
+
+params ["_unit"];
+
+_curators = allCurators apply {getAssignedCuratorUnit _x} select {!(isNull _x)};
+
+_unit getVariable ["bis_fnc_moduleremotecontrol_owner", objNull] in _curators
```
mharis001, line 22: Missing private but, I think we could use `allPlayers` command here instead.

neilzar: I don't think using `allPlayers` is appropriate here. That command could return way more than `allCurators`, my community has 50 player missions regularly and only 4 Zeus slots. I also don't see how `allPlayers` would improve it because, as far as I can see, there is no reason any other than a Zeus would open an arsenal on a remote-controlled unit.

addons/common/functions/fnc_isRemoteControlled.sqf
```diff
@@ -0,0 +1,22 @@
...
+ * [_unit] call zen_common_fnc_isRemoteControlled
+ *
+ * Public: Yes
+ */
+
+params ["_unit"];
+
+private _curators = allCurators apply {getAssignedCuratorUnit _x} select {!(isNull _x)};
+
+_unit getVariable ["bis_fnc_moduleremotecontrol_owner", objNull] in _curators
```
mharis001, line 22: Can use an `!isNull` check after #521 .

neilzar: With the current `!isNull` select on line 20 there should be no issue, unless I am missing something. There should be no `objNull` in `_curators` to cause a false positive in case the `getVariable` returns `objNull`

mharis001: I meant just: `!isNull (_unit getVariable ["bis_fnc_moduleremotecontrol_owner", objNull])` since #521 makes it so the variable is reset after disconnect.

Either is fine though really. Only suggestion would be to remove the parentheses around `isNull` on L20.

mharis001 (approved):

CreepPork (approved):

PR #534, opened 2021-01-27
Sentence case comment linter

CreepPork:
- Adds a Python linter that will lint all comments in `*.sqf`, `*.cpp` and `*.hpp` files (only `//` comments are checked)
- It will check if comments start with a lowercase letter (introduces various inconsistencies in the code)
- Ignores comments that contain various URLs (`http`/`https`)
- Update the coding guidelines to enforce that all comments start with a uppercase letter
- Fix existing inconsistencies in the code base regarding comments

For the future (will not be included in this PR):
- Add a feature that will also check if there is a space or not between the start of the comment and the comment text

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,8 @@ jobs:
       run: python3 tools/stringtable_validator.py
     - name: Validate Return Types
       run: python3 tools/return_checker.py
+    - name: Validate Comments
+      run: python3 tools/comment_linter.py
     - name: Check for BOM
       uses: arma-actions/bom-check@master
   build:
--- a/docs/development/coding_guidelines.md
+++ b/docs/development/coding_guidelines.md
@@ -307,9 +307,9 @@ All inline comments must start with a uppercase letter.
 Example:
 
 ```sqf
-//// Comment   // < incorrect
-// Comment     // < correct
-/* Comment */  // < correct
+//// Comment   // < Incorrect
+// Comment     // < Correct
+/* Comment */  // < Correct
 ```
 
 ### 5.4 Comments In Code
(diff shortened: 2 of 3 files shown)


mharis001 (changes requested):
I dislike this since it does not correctly support the following situations:

- Multi-line comments that start each line with `//`. Lines that continued the previous line's sentence are forced to start with a capital, which IMO is a worse style and less readable.
- Comments that start with command names or config entry names are forced to start with a capital, which makes their casing incorrect.

CreepPork:
> * Multi-line comments that start each line with `//`. Lines that continued the previous line's sentence are forced to start with a capital, which IMO is a worse style and less readable.

Yep, this sounds reasonable, and easily done. If comments are chained one after another, then, if the previous comment doesn't end with a `.`, you could not detect that as a error.

> * Comments that start with command names or config entry names are forced to start with a capital, which makes their casing incorrect.

For this I don't have a very good solution apart from loading everything from SQF into the script, which would make things very ugly. I'd guess it would be easier to just change the sentence structure so that the commands don't start first. If the first point would be implemented it would, for the most part, remove the requirement for this point.

mharis001:
> Yep, this sounds reasonable, and easily done. If comments are chained one after another, then, if the previous comment doesn't end with a `.`, you could not detect that as a error.

Yeah, that should work well.

> For this I don't have a very good solution apart from loading everything from SQF into the script, which would make things very ugly. I'd guess it would be easier to just change the sentence structure so that the commands don't start first. If the first point would be implemented it would, for the most part, remove the requirement for this point.

Could get the command names from the [wiki](https://community.bistudio.com/wiki/Category:Scripting_Commands) and parse them using a simple regex.

PR #547, opened 2021-02-11
Change Hide Terrain Objects module radius option to a slider

ampersand38:
- Change the radius parameter from edit box to slider 1 to 1000
- Prevents user from setting it too high and crashing their game

### DIFF
--- a/addons/modules/functions/fnc_moduleHideTerrainObjects.sqf
+++ b/addons/modules/functions/fnc_moduleHideTerrainObjects.sqf
@@ -53,5 +53,5 @@ deleteVehicle _logic;
         _objectTypes append ["ROCK", "ROCKS", "SHIPWRECK", "HIDE"];
     };
 
-    [QGVAR(hideTerrainObjects), [_position, _objectTypes, parseNumber _range, _hide]] call CBA_fnc_serverEvent;
+    [QGVAR(hideTerrainObjects), [_position, _objectTypes, _radius, _hide]] call CBA_fnc_serverEvent;
 }, {}, _position] call EFUNC(dialog,create);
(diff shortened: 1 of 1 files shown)


addons/modules/functions/fnc_moduleHideTerrainObjects.sqf
```diff
@@ -55,5 +55,5 @@ deleteVehicle _logic;
 
     {
         [QEGVAR(common,hideObjectGlobal), [_x, _hide]] call CBA_fnc_serverEvent;
-    } forEach nearestTerrainObjects [_position, _objectTypes, parseNumber _range];
+    } forEach nearestTerrainObjects [_position, _objectTypes, parseNumber _radius];
```
neilzar, line 58: `parseNumber` not needed.

neilzar (changes requested):

addons/modules/functions/fnc_moduleHideTerrainObjects.sqf
ampersand38: aadd2d2

addons/modules/functions/fnc_moduleHideTerrainObjects.sqf
```diff
@@ -26,14 +26,14 @@ deleteVehicle _logic;
         "str_3den_garbagecollection_attribute_mode_displayname",
         [true, 1, 2, ["STR_Disp_Show", "STR_Disp_Hide"]]
     ],
-    ["EDIT", ELSTRING(common,Radius_Units), 10],
+    ["SLIDER", ELSTRING(common,Radius_Units), [1, 100, 10, 0]],
```
mharis001, line 29: Maximum of 100 m seems very low,

ampersand38: That's kind of hiding a small woods area or city blocks already. I'm certainly open to increasing it to a more useful value.

CreepPork: I think 500-1000 meters would be nice.

CreepPork (approved):

addons/modules/functions/fnc_moduleHideTerrainObjects.sqf
mharis001: 1000 m is probably a good maximum. We can increase it later if needed.

mharis001 (approved):

neilzar (approved):

PR #548, opened 2021-02-11
Optimize Hide Terrain Objects module server events

mharis001:
- title, only one server event instead of one for every object


ampersand38 (approved):
That's great!

neilzar (approved):

CreepPork (approved):

PR #549, opened 2021-02-12
Add a Radius option for custom dialog sliders

neilzar:
- Title
- Draws a circle on the terrain to indicate the affected area.


neilzar:
I can't think of a reliable way to determine the module if the module coder didn't go with the `_logic` standard.

mharis001:
I think this shouldn't use `_logic` as the hardcoded variable name. Instead, the center can be specified by providing an object or AGL position in the value info. That was the dialog can support any number of radius sliders.

Could also add an optional color parameter to value info.

addons/dialog/functions/fnc_create.sqf
```diff
@@ -324,4 +340,23 @@ _display displayAddEventHandler ["KeyDown", {
...
+            private _radius = GVAR(radiuses) select _forEachIndex;
+
+            private _count = CIRCLE_DOTS_MIN max floor (2 * pi * _radius / CIRCLE_DOTS_SPACING);
+            private _factor = 360 / _count;
+
+            for "_i" from 0 to (_count - 1) do {
+                private _phi = (_i * _factor);
+                private _posVector = [_radius * cos(_phi), _radius * sin(_phi), 0];
+
+                drawIcon3d ["\A3\ui_f\data\map\markers\military\dot_CA.paa", _color, _center vectorAdd _posVector, CIRCLE_DOTS_SCALE, CIRCLE_DOTS_SCALE, 0];
```
mharis001, line 356: `drawIcon3D`.

addons/dialog/functions/fnc_create.sqf
```diff
@@ -324,4 +340,23 @@ _display displayAddEventHandler ["KeyDown", {
...
+        {
+            _x params ["_center", "_color"];
+            private _radius = GVAR(radiuses) select _forEachIndex;
+
+            private _count = CIRCLE_DOTS_MIN max floor (2 * pi * _radius / CIRCLE_DOTS_SPACING);
+            private _factor = 360 / _count;
+
+            for "_i" from 0 to (_count - 1) do {
+                private _phi = (_i * _factor);
+                private _posVector = [_radius * cos(_phi), _radius * sin(_phi), 0];
```
mharis001, line 354: No parentheses.

mharis001 (commented):
I think this would be better if the handling for the drawing was in `gui_slider`. The `Draw3D` EH could be self-cleaning and use the slider position.

addons/dialog/functions/fnc_gui_slider.sqf
```diff
@@ -22,11 +22,31 @@
...
+_settings params ["_min", "_max", "_formatting", "_isPercentage", "_drawRadius", "_radiusCenter", "_radiusColor"];
 
 private _ctrlSlider = _controlsGroup controlsGroupCtrl IDC_ROW_SLIDER;
 private _ctrlEdit = _controlsGroup controlsGroupCtrl IDC_ROW_EDIT;
 
+if (_drawRadius) then {
+    [missionNamespace, "Draw3D", {
+        _thisArgs params ["_ctrlSlider", "_center", "_color"];
+
+        if (isNull _ctrlSlider || _center isEqualTo objNull) exitWith { removeMissionEventHandler [_thisType, _thisID]; };
```
mharis001, line 34: Multi-line and lazy eval.

addons/dialog/functions/fnc_gui_slider.sqf
```diff
@@ -22,11 +22,31 @@
...
+        if (_center isEqualType objNull) then { _center = ASLToAGL getPosASL _center; };
```
mharis001, line 35: Multi-line and blank line between this `if` and the one above.

addons/dialog/functions/fnc_gui_slider.sqf
```diff
@@ -22,11 +22,36 @@
...
+if (_drawRadius) then {
+    [missionNamespace, "Draw3D", {
+        _thisArgs params ["_ctrlSlider", "_center", "_color"];
+
+        if (isNull _ctrlSlider || {_center isEqualTo objNull}) exitWith {
+            removeMissionEventHandler [_thisType, _thisID];
+        };
+
+        if (_center isEqualType objNull) then {
+            _center = ASLToAGL getPosASL _center;
```
mharis001, line 39: Probably want to use `getPosASLVisual` here for smoother drawing around moving objects.

addons/modules/functions/fnc_moduleEarthquake.sqf
```diff
@@ -18,7 +18,7 @@
 params ["_logic"];
 
 [LSTRING(ModuleEarthquake), [
-    ["SLIDER", LSTRING(ModuleEarthquake_Radius), [0, 5000, 200, 0]],
+    ["SLIDER:RADIUS", LSTRING(ModuleEarthquake_Radius), [0, 5000, 200, 0, _logic, [1, 0, 0, 0.7]]],
```
mharis001, line 21: `_logic` is deleted, need to use position here.

mharis001 (commented):
Should maybe add some scaling to the dots spacing since with larger radii a very large amount of dots are drawn. For example, with a radius of 1000 m, 1256 dots are drawn.

addons/modules/functions/fnc_moduleEarthquake.sqf
```diff
@@ -18,7 +18,7 @@
 params ["_logic"];
 
 [LSTRING(ModuleEarthquake), [
-    ["SLIDER", LSTRING(ModuleEarthquake_Radius), [0, 5000, 200, 0]],
+    ["SLIDER:RADIUS", LSTRING(ModuleEarthquake_Radius), [0, 5000, 200, 0, ASLToAGL getPosASL _logic, [1, 0, 0, 0.7]]],
```
CreepPork, line 21: Why here is position converted but other modules have the object directly passed?

neilzar: In this case, only the original position is needed; so this saves some unneeded processing during each frame. I'm not sure if this is a case where the logic is deleted before the module is applied, but passing just the logic would have it get the position of the logic every frame despite it not moving.

mharis001: Logic is deleted on line 47. Should get the position once like other modules and use that:

```sqf
private _position = ASLtoAGL getPosASL _logic;
deleteVehicle _logic;
```

docs/frameworks/dynamic_dialog.md
```diff
@@ -163,6 +163,10 @@ A percentage slider sub-type exists for this control type - `SLIDER:PERCENT`.
 When this sub-type is used, the value displayed in the edit box will be multiplied by 100 and be suffixed by a percent sign.
 The returned value will still be within the min and max values (ideally 0 to 1) and the formatting argument is ignored.
 
+A radius slider sub-type exists for this control type - `SLIDER:RADIUS`.
+When this sub-type is used, a circle is drawn on the terrain to indicate the radius of the output value. The circle is drawn around the center of the object or position in the given color.
+The circle will be drawn around the center given as argument, with the color given.
```
mharis001, line 168: This line is included in the previous line.

addons/dialog/functions/fnc_gui_slider.sqf
```diff
@@ -22,11 +22,36 @@
...
+        if (isNull _ctrlSlider || {_center isEqualTo objNull}) exitWith {
+            removeMissionEventHandler [_thisType, _thisID];
+        };
+
+        if (_center isEqualType objNull) then {
+            _center = ASLToAGL getPosASLVisual _center;
+        };
+
+        private _radius = sliderPosition _ctrlSlider;
+        private _count = CIRCLE_DOTS_MIN max floor (2 * pi * sqrt _radius / CIRCLE_DOTS_SPACING);
```
mharis001, line 43: The dots are a bit sparse now. Consider doing `_radius ^ 0.65` instead of `sqrt` (or maybe another value between 0.6 and 0.7).

mharis001 (approved):

CreepPork (approved):

PR #552, opened 2021-02-13
Convert compositions to hashes

neilzar:
- Title
- Should improve performance slightly as findif is no longer needed.


addons/compositions/functions/fnc_processTreeAdditions.sqf
```diff
@@ -38,34 +33,30 @@ for "_i" from 0 to ((_ctrlTree tvCount [0]) - 1) do {
 
             private "_path";
 
-            if (_category in _categories) then {
-                // Try to find previously added category's current path in the tree
-                for "_j" from 0 to ((_ctrlTree tvCount [0, _i]) - 1) do {
-                    if (_ctrlTree tvText [0, _i, _j] == _category) exitWith {
-                        _path = [0, _i, _j];
-                    };
```
mharis001, line 46: Why remove these lines? They handle an active search which hides tree paths.

CreepPork (approved):

mharis001:
Will check if tree additions are still processed correctly.

addons/compositions/XEH_preInit.sqf
```diff
@@ -8,4 +8,25 @@ PREP_RECOMPILE_END;
 
 GVAR(randomize) = false;
 
+private _compositions = GET_COMPOSITIONS;
+
+if (typeName _compositions isEqualTo "ARRAY") then {
```
mharis001, line 13: `isEqualType`.

addons/compositions/script_component.hpp
```diff
@@ -48,12 +48,9 @@
 #define COMPOSITION_STR QUOTE(COMPOSITION)
 
 #define VAR_COMPOSITIONS QGVAR(data)
-#define GET_COMPOSITIONS (profileNamespace getVariable [VAR_COMPOSITIONS, []])
-#define SET_COMPOSITIONS(value) (profileNamespace setVariable [VAR_COMPOSITIONS, value])
```
mharis001, line 52: I'd prefer if we kept the `SET_COMPOSITIONS` macro.

addons/compositions/XEH_preInit.sqf
```diff
@@ -8,4 +8,25 @@ PREP_RECOMPILE_END;
...
+        _x params ["_category", "_name", "_data"];
+
+        private _categoryHash = _newCompositions get _category;
+
+        if (isNil "_categoryHash") then {
+            _categoryHash = createHashMapFromArray [[_name, _data]];
+            _newCompositions set [_category, _categoryHash];
+        } else {
+            _categoryHash set [_name, _data];
+        };
```
mharis001, line 26: `getOrDefault` with `setDefault` parameter as `true` can simplify this code.

addons/compositions/functions/fnc_openDisplay.sqf
```diff
@@ -111,21 +109,31 @@ private _ctrlButtonOK = _display displayCtrl IDC_OK;
...
-        SET_COMPOSITIONS(_compositions);
+        private _categoryHash = _compositions get _category;
+
+        if (isNil "_categoryHash") then {
+            // Create category hash if category doesn't exist yet.
+            _categoryHash = createHashMapFromArray [[_name, _compositionData]];
+            _compositions set [_category, _categoryHash];
+        } else {
+            _categoryHash set [_name, _compositionData];
+        };
```
mharis001, line 127: `getOrDefault` with `setDefault` parameter as `true` can simplify this.

addons/compositions/functions/fnc_openDisplay.sqf
```diff
@@ -36,10 +36,8 @@ _ctrlList ctrlAddEventHandler ["LBSelChanged", {
...
 
 {
-    _x params ["_category"];
-
-    if (_categories pushBackUnique _category != -1) then {
-        _ctrlList lbAdd _category;
+    if (_categories pushBackUnique _x != -1) then {
+        _ctrlList lbAdd _x;
     };
 } forEach GET_COMPOSITIONS;
```
mharis001, line 42: I think we can use the `keys` command here.

addons/compositions/XEH_preInit.sqf
neilzar: This was done on purpose as a hash is also seen as an array by isEqualType, unless that has changed in a recent update.

mharis001: Looks to be working as expected now:

```sqf
createHashMap isEqualType [] // -> false
```

addons/compositions/functions/fnc_openDisplay.sqf
```diff
@@ -36,12 +36,9 @@ _ctrlList ctrlAddEventHandler ["LBSelChanged", {
...
 {
-    _x params ["_category"];
-
-    if (_categories pushBackUnique _category != -1) then {
-        _ctrlList lbAdd _category;
-    };
-} forEach GET_COMPOSITIONS;
+    _categories pushBack _x;
+    _ctrlList lbAdd _x;
+} forEach keys GET_COMPOSITIONS;
```
mharis001, line 41: Don't need `_categories` variable anymore. It was only used to prevent duplicate entries from being added.

addons/compositions/functions/fnc_openDisplay.sqf
```diff
@@ -111,21 +108,25 @@ private _ctrlButtonOK = _display displayCtrl IDC_OK;
...
     if (_mode == "create") then {
         // In create mode, add the composition to saved data
         private _compositions = GET_COMPOSITIONS;
-        _compositions pushBack _composition;
-        SET_COMPOSITIONS(_compositions);
+        private _categoryHash = _compositions getOrDefault [_category, createHashMap, true];
+
+        _categoryHash set [_name, _compositionData];
+
+        SET_COMPOSITIONS(VAR_COMPOSITIONS);
```
mharis001, line 122: Should be `_compositions` and not `VAR_COMPOSITIONS`.

addons/compositions/functions/fnc_openDisplay.sqf
```diff
@@ -111,21 +108,25 @@ private _ctrlButtonOK = _display displayCtrl IDC_OK;
...
     } else {
         // In edit mode, remove the old composition from the tree
         [false] call FUNC(removeFromTree);
     };
```
mharis001, line 126: Does this work in edit mode? I do not see where the existing composition's category and name are updated.

neilzar: In my testing, it seemed to work. I may have missed it though, was some time ago.

neilzar: The new code I pushed should solve that problem.

mharis001: Not fixed. Edit a composition by changing its name and category and reload the interface. Notice that the composition now exists under both the old and new name and category combinations.

mharis001:
Doesn't correctly delete empty category hashes. I think this likely just needs to call `removeFromTree` with `_deleteFromData` as `true` in edit mode.

CreepPork (approved):

mharis001 (commented):
Seems to work as expected now. Just some style/cleanup suggestions.

addons/compositions/functions/fnc_openDisplay.sqf
```diff
@@ -104,29 +98,31 @@ private _ctrlButtonOK = _display displayCtrl IDC_OK;
...
+    SET_COMPOSITIONS(_compositions);
+
     // Add the new/updated composition to the tree
-    GVAR(treeAdditions) pushBack +_composition;
+    GVAR(treeAdditions) pushBack [_category, _name, +_data];
     [findDisplay IDD_RSCDISPLAYCURATOR] call FUNC(processTreeAdditions);
 
     saveProfileNamespace;
-}, _this] call CBA_fnc_addBISEventHandler;
+}, [_mode, _composition select 2]] call CBA_fnc_addBISEventHandler;
```
mharis001, line 128: Add `_data` to `params` on line 90.

mharis001 (approved):

Issue #576, opened 2021-03-24
addAttribute Value Info evaluation


ffredyk:
**Is your enhancement related to a problem?**
I can't create dynamic combobox attribute, because the Value Info variable is required type Array for attributes of type zen_attributes_combo. 

**Solution you'd like:**
A small change in the file "/addons/attributes/functions/fnc_gui_combo.sqf" to also accept _valueInfo as a code {} which can then be evaluated (call) to receive dynamic array.

**Alternatives you've considered:**
Figuring out a solution involving constant direct changing of zen_attributes_displays which is an exceptionally dumb way to accomplish dynamic combo attributes, but might be the only one at the moment.

**Additional context:**
None

PR #578, opened 2021-03-24
add: _valueInfo evaluation

ffredyk:
- Add _valueInfo evaluation only if the variable is a codeblock (typeName = "CODE")
- Evaluate _valueInfo to be able to create combo attributes with dynamic values
- Fallback to previous behaviour if _valueInfo is not a codeblock

PR for #576

### DIFF
--- a/addons/attributes/functions/fnc_gui_combo.sqf
+++ b/addons/attributes/functions/fnc_gui_combo.sqf
@@ -20,7 +20,15 @@
  */
 
 params ["_controlsGroup", "_defaultValue", "_valueInfo"];
+
+//Value Info Eval
+if(typeName _valueInfo == typeName {}) then {
+    _valueInfo = call _valueInfo;
+    if(typeName _valueInfo != typeName []) then { _valueInfo = [] };
+};
+
 _valueInfo params [["_entries", [], [[]]]];
+
 
 private _ctrlCombo = _controlsGroup controlsGroupCtrl IDC_ATTRIBUTE_COMBO;
 


addons/attributes/functions/fnc_gui_combo.sqf
```diff
@@ -20,8 +20,16 @@
...
+
+//Value Info Eval
+if(typeName _valueInfo == typeName {}) then {
+    _valueInfo = call _valueInfo;
+    if(typeName _valueInfo != typeName []) then { _valueInfo = [] };
+};
+
 _valueInfo params [["_entries", [], [[]]]];
 
+
```
mharis001, line 32: I think adding `{}` to the allowed data types for `_entries` might be better. `_entries` is nested in `_valueInfo` to allow for other parameters to be easily added in the future. Something like:

```sqf
_valueInfo params [["_entries", [], [[], {}]]];

if (_entries isEqualType {}) then {
    _entries = call _entries;
};
```

ffredyk: It would be more convenient to be able to dynamically generate all valueinfo subparameters than just entries, wouldn't it? Your proposal would limit the dynamic possibilities to just one parameter instead of all.

mharis001: Yes, at that point I would expect all control types to support `_valueInfo` as code. I guess the only one that collides with currently is `QGVAR(edit)` since it expects a sanitizing function which can be provided outside of an array. However, that is easily resolved by requiring that it is placed in an array.

Let's make all attribute controls support dynamic `_valueInfo` then.

mharis001 (approved):

neilzar (approved):

Issue #577, opened 2021-03-24
Persistent Modules


ffredyk:
**Is your feature request related to a problem?**
Currently, the custom modules framework doesn't support creation of persistent modules. 

This logic prevents the module from existing ever. It cannot be overriden or changed by a mission script: 
https://github.com/zen-mod/ZEN/blob/d56beeda089454891b4c93615f55bfacf4fc338c/addons/custom_modules/functions/fnc_init.sqf#L24 

**Solution you'd like:**
Parameter driven deletion of custom modules, so I can apply my own logic to the persistent ones without making a custom module system by my own. This could hold another optional parameter for persistency:
https://github.com/zen-mod/ZEN/blob/d56beeda089454891b4c93615f55bfacf4fc338c/addons/custom_modules/functions/fnc_register.sqf#L30-L35
Then this could be changed to also get the persistency parameter and be used in condition for the deleteVehicle command:
https://github.com/zen-mod/ZEN/blob/d56beeda089454891b4c93615f55bfacf4fc338c/addons/custom_modules/functions/fnc_init.sqf#L21

**Alternatives you've considered:**
Currently, only a custom - custom module solution is the only viable solution.

**Additional context:**
None.

PR #579, opened 2021-03-24
Persistent modules

ffredyk:
- Add a new parameter to zen_custom_modules_fnc_register for module persistency.
- Add a condition inside the custom module initialization function to remove only non-persistent modules.

PR for #577

### DIFF
--- a/addons/attributes/functions/fnc_gui_combo.sqf
+++ b/addons/attributes/functions/fnc_gui_combo.sqf
@@ -20,7 +20,15 @@
  */
 
 params ["_controlsGroup", "_defaultValue", "_valueInfo"];
+
+//Value Info Eval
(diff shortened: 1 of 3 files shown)


mharis001:
One problem with this is that the module does not have the correct name, category, or icon in the entities tree.

ffredyk:
@mharis001 well I am not going to try and implement that kind of UI hacking, so if nobody else is up for the job, you could at least let it slide undercover and not list it publicly as a official supported function. This little change however, will greatly expand custom module possibilities and I personally would not mind the incorrect information inside the entities tree.

mharis001:
The module isn't passed as a parameter to the function. Also, if the user code really needs a logic object, it can create one. Need to think more about this.

PR #584, opened 2021-04-15
Use hash map for markers tree cache

mharis001:
- Use hash map for markers tree cache, needs next Arma version for updated `getOrDefault` syntax
- Use `tvSortAll` command


neilzar (approved):

addons/main/script_mod.hpp
```diff
@@ -10,7 +10,7 @@
 #define VERSION_AR  MAJOR,MINOR,PATCHLVL,BUILD
 
 // MINIMAL required version for the Mod. Components can specify others..
-#define REQUIRED_VERSION 2.02
+#define REQUIRED_VERSION 2.04
```
CreepPork, line 13: This is changed in #587. Doesn't it make sense for it there to be removed?

mharis001: Removed from this PR after merging master into this PR's branch.

CreepPork (approved):

PR #585, opened 2021-04-20
Add injury module, resolves #452

### LINKED ISSUE #452: Feature Request: Injury module (to injure players or AI)
**Problem:** ZEN does not seem to have any modules for injuring/wounding players, particularly AI.  

**Solution:**
Add a module similar to what Achilles has that instantly applies wounds/injuries to AI or player.  Types of injuries, body parts injured, severity of injuries, and consciousness of casualty would ideally be adjustable.  

**Alternatives considered:**
Only alternative I've found is continuing to use Achilles (I prefer not to have both) or manually wounding the AI (which is tedious and time consuming).

CreepPork:
- Add a module to injure players or AI
- Resolve #452 
- [ ] todo: Compatibility with ACE3

### DIFF
--- a/addons/damage/stringtable.xml
+++ b/addons/damage/stringtable.xml
@@ -655,5 +655,14 @@
             <Korean>슬랫 뒤</Korean>
             <Japanese>スラット 後部</Japanese>
         </Key>
+        <Key ID="STR_ZEN_Damage_HitHead">
+            <English>Head</English>
+        </Key>
+        <Key ID="STR_ZEN_Damage_HitHands">
+            <English>Arms</English>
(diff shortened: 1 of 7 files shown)


CreepPork:
Looking quickly through ACE3 codebase, I don't know what API functions I need to use to make it work with ACE medical. As far as I can find, the developer docs on ACE3 have nothing on the medical components.

mharis001:
You can use this [function](https://github.com/acemod/ACE3/blob/master/addons/medical/functions/fnc_addDamageToUnit.sqf) to add ACE medical injuries to units.

PR #595, opened 2021-06-11
Add Suppressive Fire module

### LINKED ISSUE #222: Feature Request: Suppresive Fire module
There is no way to make AI shoot at a building/area without any enemy targets in sight.

**Solution you'd like:**
Similar to Achilles, a module that can be placed on a group or unit making that group or unit lay down suppressive fire on a selected target.

mharis001:
- Add Suppressive Fire module
    - Goal was to create a convincing suppressive fire effect even at long range (works fairly reliably with infantry at ~1000 m and even farther for vehicles)
    - Close #222 
- Add `isUnitFFV` common function
- Add `parseMagazineDetail` common function


neilzar:
Placing multiple suppressive fire modules on the same unit/group will cause several issues:
- Only the latest target will be targeted, but the old target remains until that timer is done.
- If the first module placed lasts shorter than a later target placed, the unit will lower its weapon and move again but continues firing until the later module is also done. So it is firing at the ground or in random places once it stops moving.
- The other way around, if the last module is finished before an earlier one, the unit will continue firing in the direction it is facing until the earlier placed modules have completed. Once the first placed module is finished, it will continue with the behaviour described in the point above. Moving any earlier module's target will not cause the unit to change its direction of fire, partly because of the first point.

neilzar (changes requested):
Placing multiple suppressive fire modules on the same unit/group will cause several issues:
- Only the latest target will be targeted, but the old target remains until that timer is done.
- If the first module placed lasts shorter than a later target placed, the unit will lower its weapon and move again but continues firing until the later module is also done. So it is firing at the ground or in random places once it stops moving.
- The other way around, if the last module is finished before an earlier one, the unit will continue firing in the direction it is facing until the earlier placed modules have completed. Once the first placed module is finished, it will continue with the behaviour described in the point above. Moving any earlier module's target will not cause the unit to change its direction of fire, partly because of the first point.

mharis001:
Preventing the suppressive fire function from being called on a unit that is already suppressing should fix these issues.

jannes-io:
Does this take locality into account? Thinking about units on headless clients and such? This used to be a problem in other zeus enhancement mods (achilles/ares).

CreepPork (approved):
Looks fine, but maybe some comment line breaks could be improved (e.g. in `addons/ai/functions/fnc_suppressiveFire.sqf` `:100-103`.

I don't know what is the unwritten practice of multiline comments here, but I think making comments look like this is better:
```sqf
// My very long comment goes here
//  and here is the continuation of this, again, very long comment. 
```

neilzar (approved):
Previous issues have been solved. I only found that after applying the module to a single unit and then using the module again on the entire group, it will also draw a line for the unit that already had the module applied. 
Doesn't break functionality, it will continue with the target and settings of the previous module. But it may be a bit confusing to see a line from that unit to the new target selection as well.

mharis001:
Merging now for 1.12.0. Can look at further UX improvements in the future.

Timmy-Beal:
@mharis001 Is there a plan to has a burst fire-rate option like achilles had?

mharis001:
Not right now.

PR #604, opened 2021-06-26
Improve Tracers module weapon compatibility

mharis001:
- Use `BIS_fnc_baseWeapon` for better base weapon support
    - Some weapons do not have `baseWeapon` config entry explicitly defined
- Filter out `SAFE` muzzle (generally found in RHS weapons)

### DIFF
--- a/addons/modules/functions/fnc_compileTracers.sqf
+++ b/addons/modules/functions/fnc_compileTracers.sqf
@@ -23,11 +23,11 @@ private _cfgMagazineWells = configFile >> "CfgMagazineWells";
     if (
         getNumber (_x >> "scope") == 2
         && {getNumber (_x >> "type") == TYPE_WEAPON_PRIMARY}
-        && {count getArray (_x >> "muzzles") == 1}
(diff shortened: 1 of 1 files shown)


CreepPork (approved):
Perhaps this to 1.11.1?

jonpas:
There is also https://github.com/CBATeam/CBA_A3/blob/da4539095ca62836c555b38523918a1c36e9eb1a/addons/common/fnc_getNonPresetClass.sqf in case it would work better (uses https://github.com/CBATeam/CBA_A3/blob/da4539095ca62836c555b38523918a1c36e9eb1a/addons/common/fnc_weaponComponents.sqf for weapons). No clue which is more reliable and/or faster at this point though.

mharis001:
`BIS_fnc_baseWeapon` works well for this. This is what BI Arsenal, ACE Arsenal, and the inventory display use. Also, The CBA functions can't be used at preStart due to creating namespaces.

neilzar (approved):

PR #631, opened 2021-08-15
Tasks - Fix task type translations

mharis001:
- title, use "Attack" instead of "Assault" and "Scout" instead of "Recon"

### DIFF
--- a/addons/tasks/CfgTaskTypes.hpp
+++ b/addons/tasks/CfgTaskTypes.hpp
@@ -9,7 +9,7 @@ class CfgTaskTypes {
         displayName = "$STR_A3_cfggroups_west_blu_f_armored0";
     };
     class attack {
-        displayName = "$STR_A3_Role_Assault";
+        displayName = "$STR_A3_om_system_restui_tasktooltip";
     };
     class backpack {
         displayName = "$STR_A3_requiredgear_listitem_backpack";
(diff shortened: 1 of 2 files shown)


addons/tasks/CfgTaskTypes.hpp
```diff
@@ -9,7 +9,7 @@ class CfgTaskTypes {
         displayName = "$STR_A3_cfggroups_west_blu_f_armored0";
     };
     class attack {
-        displayName = "$STR_A3_Role_Assault";
+        displayName = "$STR_A3_om_system_restui_tasktooltip";
```
Kexanone, line 12: I'm not sure we can use translations from Old Man. I tried to avoid them so far.

addons/tasks/CfgTaskTypes.hpp
```diff
@@ -126,7 +126,7 @@ class CfgTaskTypes {
         displayName = CSTRING(Run);
     };
     class scout {
-        displayName = "$STR_A3_Role_Recon";
+        displayName = "$STR_A3_b_ctrg_soldier_tna_f0";
```
Kexanone, line 129: This entry doesn't make sense for other languages.

mharis001: Probably need to add a stringtable key then. This was the only case correct one I found.

addons/tasks/CfgTaskTypes.hpp
mharis001: Should be fine since this comes from `languagemissions_f_oldman`, which CBA requires.

Kexanone (approved):

addons/tasks/CfgTaskTypes.hpp
Kexanone: Alright

Kexanone (approved):

neilzar (approved):

PR #641, opened 2021-10-11
Update Japanese translation

classicarma:
supported of new entries


classicarma:
> Thank you for the translations as usual.

Anytime😀

PR #648, opened 2021-10-22
Add speed limit attribute

### LINKED ISSUE #605: Limit vehicle speed module
Theres one very useful command for handling AI vehicles (especially civilian or patrolling vics), and that's limitSpeed. Makes it a lot less likely for AI to ram into players and other silly stuff, and makes them drive more immersively in urban areas.

However if you want to use it on the fly through Zeus you need to type the line into each vehicles init field manually. Having this automated with a module (preferrably one that remembers the previously set limit) would make it easier to handle spawning multiple vehicles or adjusting speeds on the fly, as well as make the command more widespread among zeuses.

All the module would need to do is send `this limitSpeed XX;` into the vehicles init, replacing the X's with the set speed, and remember the previous setting. Could have a slider between 5 and 200 km/h (the high end is more for helicopters).

PS: The command does work with air and sea vehicles as well, so there's no need to limit which vehicle types it can be applied to.

ampersand38:
- Add settings to toggle certain attributes being shown (Lock, Respawn, Speed Limit)
- Add slider attribute to control vehicle speed limit
- https://youtu.be/j2aBsf6XmIE
- Close #605

### DIFF
--- a/addons/attributes/initAttributes.sqf
+++ b/addons/attributes/initAttributes.sqf
@@ -85,7 +85,7 @@
 
 [
     "Object",
-    LSTRING(SpeedLimit),
+    [LSTRING(SpeedLimit), LSTRING(SpeedLimit_Tooltip)],
     QGVAR(slider),
     [-50, 500, 5, false, 0],
     {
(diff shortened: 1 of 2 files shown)


ampersand38:
messed up the rebase/update

PR #658, opened 2021-11-28
Fix 3D rotations for Attach To module

Kexanone:
- Use `setVectorDirAndUp` rather than `setDir`

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -218,6 +218,12 @@
     _object setVectorUp _vectorUp;
 }] call CBA_fnc_addEventHandler;
 
+[QGVAR(setVectorDirAndUp), {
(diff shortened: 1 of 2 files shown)


addons/common/XEH_postInit.sqf
```diff
@@ -218,6 +218,11 @@
     _object setVectorUp _vectorUp;
 }] call CBA_fnc_addEventHandler;
 
+[QGVAR(setVectorDirAndUp), {
+    params ["_object", "_vectorDirAndUp"];
+    _object setVectorDirAndUp _vectorDirAndUp;
```
mharis001, line 223: Would prefer just `_dirAndUp`.

neilzar (approved):

mharis001 (approved):

Issue #649, opened 2021-10-24
Custom compositions not saving attached object's height propperly


ghost:
**Arma 3 Version:** `2.06` (stable)
**CBA Version:** `3.15.6.211004` (stable)
**ZEN Version:** `1.12.1` (stable)

**Mods:**
```
- CBA_A3
- ZEN
```

**Description:**
When attaching any objects, statics, vehicles etc. onto another object, static, vehicle etc.
And then saving them as a composition. When you try to spawn the composition again the attached items will not be at the correct height.

**Expected behavior:**
When I attach an object to another object and save them as a composition for later. I expect them to retain their positions when I spawn a new one as a composition.

**Steps to reproduce:**

1 - Grab a vehicle/object (here on named OBJ 1)
2 - Grab another vehicle/object (here on named OBJ 2)
3 - Disable their simulation
4 - Position OBJ 2 somewhere above OBJ 1
5 - Use the "Attach To" module 
6 - Attach OBJ 2 to OBJ 1
7 - Select OBJ 1 and OBJ 2 which are now attached
8 - Save them as a composition in the groups tab
9 - Spawn the composition

**Where did the issue occur?**
- Dedicated & Self-Hosted Multiplayer

**Additional information:**
Provided is an image of the problem.

Left is the original
Right is the spawned composition

**RPT log file:**
Unsure if its needed for this issue. If it is please reply and ill edit the issue

**Screenshots:**
![image](https://user-images.githubusercontent.com/12870917/138591668-fc5bcff2-a9a1-45d6-b296-4fae7aeb3487.png)

mharis001:
Fixed by #663.

PR #663, opened 2022-02-24
Fix attach offsets not being relative to the object's center

RedFoxIV:
https://github.com/zen-mod/ZEN/blob/36ed48447508f6020a5be609ec8da5b31624b312/addons/common/functions/fnc_serializeObjects.sqf#L200

At this line `worldToModel ASLtoAGL getPosASL` is used to get the offset required for correct attachment of _x.
The problem is, the resulting coordinate is _not_  relative to the center of the _object, but it's base. (This may be more complicated if the object is tilted, but it's still not the center.) As a result, _x is displaced vertically.

I have tested the following method for every degree of rotation along the global X axis, from 0 to 359. It seems to be a lot more accurate without any special olympics tier of crutches to negate the error in the offset.

```
_object WorldToModel ASLToAGL getPosWorld _x
```

Related issue - https://github.com/zen-mod/ZEN/issues/649


addons/common/functions/fnc_serializeObjects.sqf
```diff
@@ -197,7 +197,7 @@ private _fnc_serializeAttachedObjects = {
             private _data = _x call _fnc_serializeObject;
             if (isNil "_data") exitWith {};
 
-            private _offset = _object worldToModel ASLtoAGL getPosASL _x;
+            private _offset = _object WorldToModel ASLToAGL getPosWorld _x;
```
mharis001, line 200:
```suggestion
            private _offset = _object worldToModel ASLToAGL getPosWorld _x;
```

mharis001:
Please remove the shortcut file as well.

RedFoxIV:
Done

mharis001 (approved):

Kexanone:
Well spotted, but needs to be fixed for exported SQF as well:
https://github.com/zen-mod/ZEN/blob/7b27ef2ba30215edda21a4bbac4515dd0edb4309/addons/common/functions/fnc_exportMissionSQF.sqf#L172

Kexanone (approved):

PR #669, opened 2022-05-24
CBA Extended Loadouts

BrettMayson:
- Use CBA Extended Loadouts
- Requires https://github.com/IDI-Systems/acre2/pull/1205 to maintain ACRE radio filtering

*I think I got them all, but not tested yet*


Kexanone:
Thanks for your PR!
Definitely needs to be tested, especially with respect to backward-compatibility with already existing custom compositions, which in principle should be the case from what I've seen.
For now we have to wait for ACRE to update anyway.

addons/common/functions/fnc_getUnitLoadout.sqf
```diff
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
```
mharis001, line 1: Probably just get rid of this function and use `CBA_fnc_getLoadout` directly in other places.

addons/common/functions/fnc_hasDefaultInventory.sqf
```diff
@@ -20,7 +20,7 @@ params ["_object"];
 
 // Compare current loadout to config defined loadout if an infantry unit is given
 if (_object isKindOf "CAManBase") exitWith {
-    private _current = [_object] call FUNC(getUnitLoadout);
+    private _current = ([_object] call FUNC(getUnitLoadout)) select 0;
```
mharis001, line 23: No parentheses on LHS of `select`.

addons/context_actions/CfgContext.hpp
```diff
@@ -236,13 +236,13 @@ class EGVAR(context_menu,actions) {
         };
         class Copy {
             displayName = "$STR_3DEN_Display3DEN_MenuBar_EntityCopy_text";
-            statement = QUOTE(GVAR(loadout) = _hoveredEntity call EFUNC(common,getUnitLoadout));
+            statement = QUOTE(GVAR(extendedLoadout) = _hoveredEntity call EFUNC(common,getUnitLoadout));
```
mharis001, line 239: Would prefer leaving it as `GVAR(loadout)`.

addons/common/functions/fnc_getUnitLoadout.sqf
BrettMayson: Went to do that but wasn't sure if you'd want to keep the function around for other uses. I'll remove it 👍

mharis001:
Should bump `REQUIRED_CBA_VERSION`. Will wait on CBA update that includes fix before merging.

mharis001 (approved):

Kexanone (approved):

PR #684, opened 2022-10-01
Add mod icons to create unit trees

### LINKED ISSUE #207: Interface Symbol
When using in conjunction with multiple other zeus interface mods it can be difficult to differentiate what module comes from which mod.

Would it be possible for a symbol representing the mod to be added next to modules from this mod. Similar to how Achilles does their spartan helmet?

mharis001:
- title, Close #207

### DIFF
--- a/addons/editor/XEH_PREP.hpp
+++ b/addons/editor/XEH_PREP.hpp
@@ -1,4 +1,5 @@
 PREP(addGroupIcons);
+PREP(addModIcons);
 PREP(declutterEmptyTree);
 PREP(fixSideButtons);
 PREP(handleKeyDown);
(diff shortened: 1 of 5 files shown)


neilzar (approved):

Kexanone (changes requested):
Looks fine, but changing the setting should trigger reloading the module tree.

mharis001:
ZEN is sort of split on that right now. Some settings reload the display automatically (e.g. faction filter) while others apply once the display is manually reloaded (e.g. the similar group icons setting). I guess if we change it here, the behaviour should be made consistent across all features.

Kexanone:
> ZEN is sort of split on that right now. Some settings reload the display automatically (e.g. faction filter) while others apply once the display is manually reloaded (e.g. the similar group icons setting). I guess if we change it here, the behaviour should be made consistent across all features.

Agreed, we should have a dedicated PR for that. In fact I was about to make the same comment for the placement preview, but thought it was out of scope for that PR.

Kexanone (approved):

PR #691, opened 2022-11-15
Add attach/detach keybinds and allow editing attached objects directly

mharis001:
- Add "Attach Objects" and "Detach Objects" keybinds to attach/detach multiple objects at a time
- Allow editing attached objects directly
    - Previously, attached objects would snap back to their original position when edited through Zeus
    - Now, we manually re-attach them based on their new relative position to the parent object
- Move all attached objects handling to new `attached_objects` component


Kexanone:
Any reason you chose <kbd>Alt</kbd> over <kbd>Ctrl</kbd> for the default keybind?

Kexanone (approved):

mharis001:
Do not remember why. Would <kbd>Ctrl</kbd> be a better default?

Kexanone:
I messed up before... I meant to replace <kbd>Shift</kbd> with <kbd>Ctrl</kbd>, since I find <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>Key</kbd> a more natural choice than <kbd>Shift</kbd>+<kbd>Alt</kbd>+<kbd>Key</kbd>, but it's probably more a matter of personal preference.

mharis001:
Changed to CTRL + SHIFT + _. Feels like the most natural to use.

Kexanone:
<kbd>Ctrl</kbd>+<kbd>Shift</kbd> is actually even better. Anyway ppl can adjust it to whatever they want anyway, so no big deal in the first place.

PR #699, opened 2023-01-10
Add settings to toggle attributes and speed limit attribute

### LINKED ISSUE #605: Limit vehicle speed module

ampersand38:
- Add settings to toggle certain attributes being shown (Lock, Respawn, Speed Limit)
- Add slider attribute to control vehicle speed limit
- https://youtu.be/j2aBsf6XmIE
- Close #605

I messed up #648 and a new pr seemed like the easiest way to fix it.

### DIFF
--- a/addons/attributes/XEH_preInit.sqf
+++ b/addons/attributes/XEH_preInit.sqf
@@ -23,5 +23,6 @@ GVAR(previousMarkerColors) = [] call CBA_fnc_createNamespace;
 
 // Initialize the core/default attributes
 #include "initAttributes.sqf"
+#include "initSettings.sqf"
 
 ADDON = true;
--- a/addons/attributes/initAttributes.sqf
+++ b/addons/attributes/initAttributes.sqf
@@ -85,6 +85,21 @@
 
 [
     "Object",
+    [LSTRING(SpeedLimit), LSTRING(SpeedLimit_Tooltip)],
+    QGVAR(slider),
+    [-50, 500, 5, false, 0],
+    {
+        {
+            [QEGVAR(common,limitSpeed), [_entity, _value], _entity] call CBA_fnc_targetEvent;
+            _entity setVariable [QGVAR(SpeedLimit), _value, true];
+        } forEach call EFUNC(common,getSelectedVehicles);
+    },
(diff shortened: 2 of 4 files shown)


addons/common/functions/fnc_setSpeed.sqf
```diff
@@ -0,0 +1,60 @@
...
+#define SPEED_FAST 1
+#define SPEED_NORMAL 2
+#define SPEED_SLOW 3
+#define SPEED_BACK 4
+
+params ["_vehicle", "_speed"];
+
+if (_vehicle isEqualType []) exitWith {
+    {
+        [_x, _speed] call FUNC(setSpeed);
```
rautamiekka, line 42: I wonder if `spawn` would speed things along ?

^ Granted, my laptop doesn't run both the game and the Server instance simultaneously especially fast, and on it I noticed that `forEach`+`spawn` takes a noticeable time to get things done compared to `call`, like `CBA_fnc_setLoadout` can take something like 400-600ms (looks like it) to get the squad equipped instead of next to instantly.

^ Might help when someone has a less good Internet connection, dunno.

ampersand38: Oh interesting, so each one can run "in parallel"?
As for network, I tried to structure the CBA event and the function such that the selected array is pre-filtered and then only sent on network once, if I'm getting your meaning correctly.

PR #700, opened 2023-01-14
Add Advanced Rappelling compatibility

Kexanone:
- Add compatibility addon for using [Duda's Advanced Rappelling](https://github.com/sethduda/AdvancedRappelling) for fast-roping waypoint and reinforcements.
- Slightly refactor fastroping in main addon


Kexanone:
To Do: Get reinforcements working

mharis001:
Maybe these should be two separate waypoints (Fastrope and Advanced Rappel) because the mods could be run together. We could add conditions to waypoints so that they only show if the required mods are loaded.

Kexanone:
Sounds like a good alternative approach. Would you propose the AR waypoint in a compat. or part of the main addon? If I remember correctly we decided against having anything but ACE as soft dependency, that's why I went the compat. route in the first place.

mharis001:
Yeah, I think we can revisit that decision for some other big mods. I think both a new component (separation of features/responsibilities) or just in the `ai` component are good options. With waypoint conditions, this feature could be added without a hard dependency.

mharis001:
I separated out the waypoints into "Fastrope" and "Rappel" and reworked how reinforcements handles the different insertion methods for air vehicles. I think we should keep the Advanced Rapelling compatibility as a separate optional component.

mharis001 (approved):

mharis001:
@Kexanone, I think this is good to go. Please look over my changes and merge when ready.

Kexanone:
Looks good to me, but still needs to be tested.

mharis001:
I did the usual testing, and everything seemed okay. I think we are good to merge.

PR #703, opened 2023-01-21
Add ability to rotate placement preview objects

ampersand38:
- Enable rotation of placement preview when holding Zeus rotation modifier, default L Shift
- Preview rotation is preserved if Shift is released
- Object can be placed while Shift is held down
- Object rotation matches preview

### DIFF
--- a/addons/placement/functions/fnc_handleObjectPlaced.sqf
+++ b/addons/placement/functions/fnc_handleObjectPlaced.sqf
@@ -32,10 +32,10 @@ _object allowDamage false;
 // Apply the preview object's position to the placed object after a frame
 // Helps in preventing the object from being destroyed by being moved
 [{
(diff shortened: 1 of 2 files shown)


mharis001:
Has weird handling when the mouse world position is "past" the object's position (see images).

<details>
<summary>Images</summary>

Placement preview (cursor is below the object in both cases as seen by the icon here):

image

Normal Zeus rotation:

image

</details>

mharis001:
I pushed a simple change that should improve the behaviour and cover all cases. Could use a second look.

mharis001 (approved):

ampersand38:
> I pushed a simple change that should improve the behaviour and cover all cases. Could use a second look.

You've managed to copy the vanilla zeus rotation behaviour xD I had always wondered why it didn't point directly to the cursor, but felt a little off. It's working from 2D screenpos!

PR #709, opened 2023-02-15
Add CQB mode to Move camera to selected entity key

ampersand38:
- Add alternate mode for Move camera to selected entity key
- CQB mode moves camera to top-rear of object bounding box Useful when unit is inside multi-floor structure
- Repeated keypresses switch between modes
- https://youtu.be/TQgj4HmbUnw

### DIFF
--- a/addons/editor/functions/fnc_handleCuratorMoveCamTo.sqf
+++ b/addons/editor/functions/fnc_handleCuratorMoveCamTo.sqf
@@ -1,0 +1,31 @@
+#include "script_component.hpp"
+/*
+ * Author: Ampersand
+ * Handles the behaviour of the curatorMoveCamTo user action.
+ *
+ * Arguments:
+ * None
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call zen_editor_fnc_handleCuratorMoveCamTo
+ *
+ * Public: No
+ */
+
+private _selectedObjects = curatorSelected select 0;
+if (count _selectedObjects == 0) exitWith {};
+
+private _selectedObject = _selectedObjects select 0;
+if (isNil QGVAR(curatorMovedCamTo) || {_selectedObject != GVAR(curatorMovedCamTo)}) exitWith {
+    GVAR(curatorMovedCamTo) = _selectedObject;
+};
+
+// Second activation of curatorMoveCamTo on same object
+(0 boundingBoxReal GVAR(curatorMovedCamTo)) params ["_p1", "_p2"];
+// Move camera to top-rear of object bounding box for CQB view
+curatorCamera setPosASL (GVAR(curatorMovedCamTo) modelToWorldVisualWorld [0, _p1 select 1, _p2 select 2]);
+curatorCamera setDir getDir GVAR(curatorMovedCamTo);
+GVAR(curatorMovedCamTo) = nil;
--- a/addons/editor/XEH_PREP.hpp
+++ b/addons/editor/XEH_PREP.hpp
@@ -2,6 +2,7 @@ PREP(addGroupIcons);
 PREP(addModIcons);
 PREP(declutterEmptyTree);
 PREP(fixSideButtons);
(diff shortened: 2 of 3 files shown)


mharis001:
Looks like the base game keybind behaviour is still visible and there is a "flicker" before the camera is moved to the CQB view. Ideally, we should override the behaviour of the base game keybind.

addons/editor/functions/fnc_handleCuratorMoveCamTo.sqf
```diff
@@ -0,0 +1,31 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call zen_editor_fnc_handleCuratorMoveCamTo
+ *
+ * Public: No
+ */
+
+private _selectedObjects = curatorSelected select 0;
```
mharis001, line 18: Use `SELECTED_OBJECTS` macro.

addons/editor/functions/fnc_handleCuratorMoveCamTo.sqf
```diff
@@ -0,0 +1,31 @@
...
+if (count _selectedObjects == 0) exitWith {};
+
+private _selectedObject = _selectedObjects select 0;
+if (isNil QGVAR(curatorMovedCamTo) || {_selectedObject != GVAR(curatorMovedCamTo)}) exitWith {
+    GVAR(curatorMovedCamTo) = _selectedObject;
+};
+
+// Second activation of curatorMoveCamTo on same object
+(0 boundingBoxReal GVAR(curatorMovedCamTo)) params ["_p1", "_p2"];
+// Move camera to top-rear of object bounding box for CQB view
```
mharis001, line 28: Move up and combine with the other comment to make it easier to read.

ampersand38:
> Looks like the base game keybind behaviour is still visible and there is a "flicker" before the camera is moved to the CQB view. Ideally, we should override the behaviour of the base game keybind.

I can add a keybind set to block the base game key. The base game behaviour is kind of weird, it looks up at flying aircraft which is totally useless, and ends up toggling between angles on subsequent presses.
For the "far" view I would want to do something like 20 m above and behind, and centered on the object.

What do you think?

addons/editor/initKeybinds.sqf
```diff
@@ -75,6 +75,12 @@
     };
 }, {}, [DIK_F, [false, true, true]]] call CBA_fnc_addKeybind; // Default: CTRL + ALT + F
 
+[ELSTRING(main,DisplayName), QGVAR(curatorMoveCamTo), ["str_usract_curator_move_cam", "str_controls_tooltips_curator_move_cam"], {
+    if (!isNull curatorCamera) then {
+        [] call FUNC(handleCuratorMoveCamTo)
+    };
+}, {}, [DIK_F, [false, false, false]]] call CBA_fnc_addKeybind; // Default: F
```
mharis001, line 82: I dislike adding a separate keybind. Let's use the KeyDown EH to override the vanilla keybind. If we want to make this optional, could add a setting like "Move Camera To Behaviour" - maybe 3 options: default, CQB mode, and CQB + alternative far view.

ampersand38: ah ok. Should this code go in `camera` instead of `editor`?

mharis001: I think `editor`.

ampersand38: Does the setting belong under camera?

mharis001: Good point. Normally, all the keybind handling is in `editor` component but I guess with the setting, it would make more sense to be under `camera`. Not really sure what option is the best.

ampersand38: For now I've implemented everything in `editor`, with the setting using the `camera` category.

ampersand38:
Just noticed current method breaks vanilla lock camera to object =(

PR #711, opened 2023-03-08
Add Toggle Object Editability keybind

mharis001:
- title, use case: sometimes you only want to edit a single object that is currently not editable and instead of going through the regular editable objects dialog with a small radius, this keybind provides a quick way to make the object editable (or not) by hovering over it
- Add `updateEditableObjects` objects common function
    - Replaces the old `addObjects`/`removeObjects` common events
    - Consolidates editable object updating in the code base to a single function


ampersand38 (approved):
Works very well! I put my keybind as Alt + Right Click and am loving it.

PR #712, opened 2023-03-14
Fix laser cycling too quickly

ampersand38:
- Laserdesignator_mounted fires but cycles too quickly to reliably turn to desired state
- Add firing delay time check
- https://www.youtube.com/watch?v=LMWf6TOveJE

### DIFF
--- a/addons/common/functions/fnc_canFire.sqf
+++ b/addons/common/functions/fnc_canFire.sqf
@@ -24,6 +24,7 @@ private _unit = _unit call FUNC(getEffectiveGunner);
 alive _unit
 && {!isPlayer _unit}
 && {lifeState _unit in ["HEALTHY", "INJURED"]}
+&& {CBA_missionTime >= _unit getVariable [QGVAR(nextFireTime), 0]}
 && {
     private _vehicle = vehicle _unit;
 
@@ -39,6 +40,5 @@ alive _unit
         && {!("fake" in toLower _weapon)}
         && {_ignoreAmmo || {_ammoCount > 0} || {_weapon isKindOf ["CarHorn", configFile >> "CfgWeapons"]}}
         && {_ignoreReload || {!([_vehicle, _turretPath] call FUNC(isReloading))}}
-        && {CBA_missionTime >= _unit getVariable [QGVAR(nextFireTime), 0]}
     };
 }
--- a/addons/common/functions/fnc_canFire.sqf
+++ b/addons/common/functions/fnc_canFire.sqf
@@ -38,7 +38,7 @@ alive _unit
         _weapon != ""
         && {!("fake" in toLower _weapon)}
         && {_ignoreAmmo || {_ammoCount > 0} || {_weapon isKindOf ["CarHorn", configFile >> "CfgWeapons"]}}
-        && {CBA_missionTime >= _unit getVariable [QGVAR(nextFireTime), 0]}
         && {_ignoreReload || {!([_vehicle, _turretPath] call FUNC(isReloading))}}
+        && {CBA_missionTime >= _unit getVariable [QGVAR(nextFireTime), 0]}
     };
 }
--- a/addons/common/functions/fnc_canFire.sqf
+++ b/addons/common/functions/fnc_canFire.sqf
@@ -38,7 +38,7 @@ alive _unit
         _weapon != ""
         && {!("fake" in toLower _weapon)}
         && {_ignoreAmmo || {_ammoCount > 0} || {_weapon isKindOf ["CarHorn", configFile >> "CfgWeapons"]}}
-        && {!(_weapon isKindOf ["Laserdesignator_mounted", configFile >> "CfgWeapons"] && {CBA_missionTime < _unit getVariable [QGVAR(nextFireTime), 0]})}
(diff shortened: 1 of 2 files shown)


addons/common/functions/fnc_fireWeapon.sqf
```diff
@@ -71,6 +71,10 @@ switch (true) do {
             _unit setAmmo [_muzzle, 1e6];
         };
 
+        if (_weapon isKindOf ["Laserdesignator_mounted", configFile >> "CfgWeapons"]) then {
```
mharis001, line 74: Is there a way to detect laser weapons in general (instead of relying on inheritance from this class)?

addons/common/functions/fnc_canFire.sqf
```diff
@@ -38,6 +38,7 @@ alive _unit
         _weapon != ""
         && {!("fake" in toLower _weapon)}
         && {_ignoreAmmo || {_ammoCount > 0} || {_weapon isKindOf ["CarHorn", configFile >> "CfgWeapons"]}}
+        && {!(_weapon isKindOf ["Laserdesignator_mounted", configFile >> "CfgWeapons"] && {CBA_missionTime < _unit getVariable [QGVAR(nextFireTime), 0]})}
```
mharis001, line 41:
```suggestion
        && {CBA_missionTime >= _unit getVariable [QGVAR(nextFireTime), 0]}
```
Would work the same and support future setting of a manual lockout for any weapon/unit. Also, move below the reload check.

addons/common/functions/fnc_fireWeapon.sqf
ampersand38: We can check magazines https://github.com/zen-mod/ZEN/blob/5a4ccf6ddc0e78b079417f4992f0715c59927281/addons/common/script_component.hpp#L92

 or ammo https://github.com/zen-mod/ZEN/blob/bd5c43823cbb2dfe0f233cfedd1b4f2bdd254f63/addons/context_actions/functions/fnc_getVehicleWeaponActions.sqf#L18

mharis001 (approved):

mharis001:
@ampersand38 I changed it to check for ammo simulation which I think will be more generalized. Please merge if the changes look good to you.

ampersand38:
For high rof weapons, do we want to cache the magazine to avoid having config look up every round?

rautamiekka:
I don't see why not always cache it. Or many other data, for that matter, especially if always local to the Server.

addons/common/functions/fnc_fireWeapon.sqf
```diff
@@ -61,12 +63,20 @@ switch (true) do {
...
 
         if (_weapon isKindOf ["CarHorn", configFile >> "CfgWeapons"]) exitWith {
             _unit forceWeaponFire [_muzzle, _fireMode];
         };
 
+        // Prevent laser weapons from firing too quickly to allow for reliably switching to the desired state
+        private _ammo = getText (configFile >> "CfgMagazines" >> _magazine >> "ammo");
+        private _ammoSimulation = getText (configFile >> "CfgAmmo" >> _ammo >> "simulation");
+
+        if (_ammoSimulation == "laserDesignate") then {
```
ampersand38, line 76:
```suggestion
        private _isLaser = GVAR(laserMagazines) getOrDefault [_magazine, nil];

        if (isNil "_isLaser") then {
            private _ammo = getText (configFile >> "CfgMagazines" >> _magazine >> "ammo");
            _isLaser = getText (configFile >> "CfgAmmo" >> _ammo >> "simulation") == "laserDesignate";
        };

        if (_isLaser) then {

```

mharis001:
I don't think it is really needed (0.0011 ms improvement). Does not seem like the slow part of the function when testing with the laser designator turret.
```
Result:
0.0021 ms

Cycles:
10000/10000

Code:
private _magazine = "Laserbatteries";
private _ammo = getText (configFile >> "CfgMagazines" >> _magazine >> "ammo"); 
private _ammoSimulation = getText (configFile >> "CfgAmmo" >> _ammo >> "simulation");
_ammoSimulation == "laserDesignate"
```
```
Result:
0.001 ms

Cycles:
10000/10000

Code:
private _magazine = "Laserbatteries";
private _ammoSimulation = a getOrDefaultCall [_magazine, {
    private _ammo = getText (configFile >> "CfgMagazines" >> _magazine >> "ammo"); 
    getText (configFile >> "CfgAmmo" >> _ammo >> "simulation")
}, true];

_ammoSimulation == "laserDesignate"
```
```
Result:
0.0269 ms

Cycles:
10000/10000

Code:
a call zen_common_fnc_fireWeapon
```

addons/common/functions/fnc_fireWeapon.sqf
ampersand38: and add to XEH_preInit.sqf
```
GVAR(laserMagazines) = createHashMap;
```

ampersand38:
I meant for when the weapon is a machinegun or autocannon, but yeah it's probably fine.

rautamiekka:
> 0.0021 ms
> [...]
> 0.001 ms

It's true that it ain't exactly very much, but since it only has a pro and no cons (AFAICT), and slowdowns eventually stack ...

PR #717, opened 2023-03-18
switch to launcher

ampersand38:
- Context action to make unit switch to launcher
- Unit is frozen via disableAI "ANIM", no movement or aiming
- Unit can do certain things like throw grenades via context action
- Switching to another weapon enables ANIM again
- https://youtu.be/54ezolpkDxw

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -176,6 +176,12 @@
 [QGVAR(selectWeapon), {
     params ["_unit", "_muzzle"];
     _unit selectWeapon _muzzle;
+    if (_muzzle == secondaryWeapon _unit) then {
+        _unit playMove "amovpercmstpsraswlnrdnon";
+        _unit disableAI "ANIM";
+    } else {
+        _unit enableAI "ANIM";
+    };
 }] call CBA_fnc_addEventHandler;
 
 [QGVAR(selectWeaponTurret), {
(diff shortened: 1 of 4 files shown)


addons/common/XEH_postInit.sqf
```diff
@@ -176,6 +176,12 @@
 [QGVAR(selectWeapon), {
     params ["_unit", "_muzzle"];
     _unit selectWeapon _muzzle;
+    if (_muzzle == secondaryWeapon _unit) then {
+        _unit playMove "amovpercmstpsraswlnrdnon";
+        _unit disableAI "ANIM";
+    } else {
+        _unit enableAI "ANIM";
+    };
```
mharis001, line 184: I think this is a decent workaround. However, IMO would be better placed in the `switchWeapon` function itself (since it is specific to that).

ampersand38: switchWeapon function is local to the curator, and these need to be local to the AI. should I make a separate selectWeaponCurator event?

addons/common/XEH_postInit.sqf
```diff
@@ -141,6 +141,11 @@
 [QGVAR(doWatch), {
     params ["_unit", "_target"];
     _unit doWatch _target;
+    {
+        if (!(_x checkAIFeature "ANIM")) then {
```
rautamiekka, line 145:
```suggestion
        if !(_x checkAIFeature "ANIM") then {
```

rautamiekka (changes requested):

PR #720, opened 2023-03-20
Improve Watch Cursor and Watch Curator Camera keybinds

mharis001:
- title, locks vehicle turret cameras to the target to make the watch command have better targeting
- Also use `doTarget` when the target is given as an object (instead of position) - makes infantry raise their weapons without needing to be in combat mode


ampersand38 (approved):
this is awesome! https://youtu.be/Tg_GpqGRUAA

PR #721, opened 2023-03-21
Add Toggle Laser keybind

mharis001:
- title, pairs nicely with the force fire keybind
- Useful because we don't list laser weapons in the context menu action for switching vehicle weapons and this would be much faster than switching back and forth between laser and missiles (for example)

### DIFF
--- a/addons/editor/stringtable.xml
+++ b/addons/editor/stringtable.xml
@@ -393,6 +393,12 @@
         <Key ID="STR_ZEN_Editor_ForceFire_Description">
             <English>Makes selected AI units fire their current weapon while the key is held down. Vehicles will fire their first available turret with a weapon.</English>
         </Key>
+        <Key ID="STR_ZEN_Editor_ToggleLaser">
+            <English>Toggle Laser</English>
+        </Key>
+        <Key ID="STR_ZEN_Editor_ToggleLaser_Description">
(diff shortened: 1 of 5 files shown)


addons/editor/stringtable.xml
```diff
@@ -393,6 +393,12 @@
         <Key ID="STR_ZEN_Editor_ForceFire_Description">
             <English>Makes selected AI units fire their current weapon while the key is held down. Vehicles will fire their first available turret with a weapon.</English>
         </Key>
+        <Key ID="STR_ZEN_Editor_ToggleLaser">
+            <English>Toggle Laser</English>
+        </Key>
+        <Key ID="STR_ZEN_Editor_ToggleLaser_Description">
+            <English>Makes selected AI units (vehicles) toggle their turret laser weapons on or off.</English>
```
rautamiekka, line 400: IMO clearer like this IIUC:
```suggestion
            <English>Makes selected AI units in vehicles toggle their turret laser weapons on or off.</English>
```
I was gonna suggest `AI vehicles` instead, but remembered that drones exist (UGV Stomper and MQ-12 Falcon FTW), which are internally using an AI unit cuz 'Arma' Engine.

ampersand38 (approved):
Nice one. Should we auto-add all laserTargets to Zeus editable?

mharis001:
No, i don't think you would ever want to edit those directly.

PR #729, opened 2023-04-05
Add Ping Zeus keybind

### LINKED ISSUE #728: Zeus ping other zeuses
While zeusing with mutiple people it gets complicated to quickly point out a location on the map.
The usual conversation goes like this: 
_"We should probably place down a MG nest around here"_
"Where's here?"
_"Right next to the town of Generic on the highpoint north of it"_
"Let me find that town...Ah! That high point, placed it!"
_"No... I meant a different highpoint."_

**Solution you'd like:**
I propose a function similar to the ping zeus feature but it should only ping *other zeuses* (and only the ones currently in the zeus interface).
The "zeus ping" will manifest the same way as the usual ping with a blib on the map.
If the zeus presses the "jump to ping" keybind their camera will be moved into the exact position and orientation as the zeus which pinged originally.

Calling out grids. Moving some placeholder object like a football there so it's easier to find.

**Additional context:**
If the zeus pings while having the map open ping the location of the map where the cursor was at that point.

ampersand38:
- Add keybind to Ping other Curators on the selected object or mouse position
- Receiving curator will get ping and icon hint, can jump to object
- https://youtu.be/HMo-H6wfMEc
- Close #728


rautamiekka:
Mmmm, I like the sound of that even without the vid.

Kexanone:
Could be bound to <kbd>U</kbd> by default, which is used in Reforger.

addons/editor/initKeybinds.sqf
```diff
@@ -113,6 +113,17 @@
     };
 }, {}, [DIK_R, [true, true, false]]] call CBA_fnc_addKeybind; // Default: CTRL + SHIFT + R
 
+[ELSTRING(main,DisplayName), QGVAR(pingCurators), [LSTRING(pingCurators), LSTRING(pingCurators_Description)], {
+    if (!isNull curatorCamera && {!GETMVAR(RscDisplayCurator_search,false)}) then {
+        private _thisCurator = getAssignedCuratorLogic player;
+        private _object = (SELECTED_OBJECTS + [_thisCurator]) select 0;
+        if (_object == _thisCurator) then {_thisCurator setPosASL ([] call EFUNC(common,getPosFromScreen))};
+        [QGVAR(pingCurator), [_object], allCurators - [_thisCurator]] call CBA_fnc_targetEvent;
```
Kexanone, line 121: I think it's better when the curator that initiates the ping also sees the ping.

Kexanone (changes requested):

addons/editor/initKeybinds.sqf
ampersand38: I see the advantage to that for sure.

addons/editor/XEH_postInit.sqf
```diff
@@ -3,3 +3,14 @@
 ["zen_curatorDisplayLoaded", LINKFUNC(handleLoad)] call CBA_fnc_addEventHandler;
 ["zen_curatorDisplayUnloaded", LINKFUNC(handleUnload)] call CBA_fnc_addEventHandler;
 [QGVAR(modeChanged), LINKFUNC(fixSideButtons)] call CBA_fnc_addEventHandler;
+
+[QGVAR(pingCurator), {
+    params ["_object"];
+
+    private _curatorLogic = getAssignedCuratorLogic player;
+    if (isNull _curatorLogic) exitWith {};
```
mharis001, line 11: I think `isNull curatorCamera` might be better here.

addons/editor/XEH_postInit.sqf
```diff
@@ -3,3 +3,14 @@
...
+    [_curatorLogic, _object] call BIS_fnc_curatorPinged;
+    [[
+        ["ICON", [_object, "\a3\ui_f\data\GUI\Cfg\Hints\Zeus_ca.paa"]]
```
mharis001, line 14: I think this should use the Zeus eye icon (not squared off one).

mharis001 (changes requested):

addons/editor/initKeybinds.sqf
```diff
@@ -113,6 +113,17 @@
     };
 }, {}, [DIK_R, [true, true, false]]] call CBA_fnc_addKeybind; // Default: CTRL + SHIFT + R
 
+[ELSTRING(main,DisplayName), QGVAR(pingCurators), [LSTRING(pingCurators), LSTRING(pingCurators_Description)], {
+    if (!isNull curatorCamera && {!GETMVAR(RscDisplayCurator_search,false)}) then {
+        private _thisCurator = getAssignedCuratorLogic player;
+        private _object = (SELECTED_OBJECTS + [_thisCurator]) select 0;
+        if (_object == _thisCurator) then {_thisCurator setPosASL ([] call EFUNC(common,getPosFromScreen))};
```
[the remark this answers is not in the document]
ampersand38: How about creating a game logic?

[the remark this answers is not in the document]
mharis001: I think that would be worse. I am not sure what BIS_fnc_curatorPinged` does but maybe we can just expand on its functionality to support pinging objects and positions.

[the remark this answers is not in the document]
ampersand38:
```sqf
/*
	Author: Karel Moricky

	Description:
	Show effects when player pings curator.
	Icon and entity list highlighting is handled by engine.

	Parameter(s):
		0: OBJECT - curator module
		1: OBJECT - player

	Returns:
	BOOL
*/

//--- Terminate when not enough thime passed since the last ping
private ["_time"];
_time = missionnamespace getvariable ["bis_fnc_curatorPinged_time",0];
if (time - _time < 0.1) exitwith {};
missionnamespace setvariable ["bis_fnc_curatorPinged_time",time];

//--- Get arguments
private ["_curator","_player","_pitch","_source","_soundPing","_sound","_volume","_pitchBase"];
_curator = _this param [0,objnull,[objnull]];
_player = _this param [1,objnull,[objnull]];
_isCurator = getassignedcuratorunit _curator == player;

//--- Mark position of the last ping (for external use)
missionnamespace setvariable ["bis_fnc_curatorPinged_player",_player];

//--- Play pinging sound
_soundsPing = getarray (configfile >> "cfgCurator" >> "soundsPing");
if (count _soundsPing > 0 || !_isCurator) then {
	_sound = "";
	if (_isCurator) then {
		//--- Curator - assign unique sound to each player
		_n = 0;
		{_n = _n + _x;} foreach (toarray name _player);
		_sound = _soundsPing select (_n % (count _soundsPing))
	} else {
		//--- Store that player already pinged, so hint is not displayed anymore
		if !(isnil {missionnamespace getvariable "bis_fnc_curatorPinged_index"}) then {
			profilenamespace setvariable ["bis_fnc_curatorPinged_done",true];
			saveprofilenamespace;
		};

		//--- Player - pick random sound to create a melody (ever played Joruney? :)
		_soundsPingCount = count _soundsPing;
		_index = missionnamespace getvariable ["bis_fnc_curatorPinged_index",floor random _soundsPingCount];
		_index = round ((_index - 2 + random 4) + _soundsPingCount) % _soundsPingCount;
		_sound = _soundsPing select _index;
		missionnamespace setvariable ["bis_fnc_curatorPinged_index",_index];
	};
	playsound [_sound,true];
};

//--- Show animated icon in the middle of the screen for player
if !(_isCurator) then {
	([] call bis_fnc_rscLayer) cutrsc ["RscCuratorPing","plain"];
};

true
```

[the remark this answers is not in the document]
ampersand38: then used here
```sqf
// P:\a3\ui_f_curator\UI\Displays\RscDisplayCurator.sqf
case (_key in actionkeys 'curatorPingView'): { //--- Go to last ping
    _target = missionnamespace getvariable "bis_fnc_curatorPinged_player";
    if !(isnil "_target") then {
        _pos = getposatl _target;
        _posZ = (getposatl curatorcamera select 2) max 5;
        _pos set [2,(_pos select 2) + _posZ];
        _pos = [_pos,_posZ * 2,direction curatorcamera + 180] call bis_fnc_relpos;
        [_pos,_target] spawn bis_fnc_setcuratorcamera;
    };
    _return = true;
};
```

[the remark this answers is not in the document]
ampersand38: > I dislike the side effect of moving the curator logic object. I guess if we want to avoid that, we would need to add our own ping system.

I mean, if not this, what else could the location of the curator logic object be used for?

addons/editor/stringtable.xml
```diff
@@ -411,5 +411,13 @@
         <Key ID="STR_ZEN_Editor_ToggleAIPath_Description">
             <English>Makes selected AI units start or stop moving. Does not affect aiming or shooting. Similar to "Hold Position" in RTS games.</English>
         </Key>
+        <Key ID="STR_ZEN_Editor_pingCurators">
+            <English>Ping Zeus</English>
```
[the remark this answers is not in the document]
Kexanone: The reason I switched to Zeus is that it is more consistent with the localization in the base game. Curator is more of an internal name, but in the manuals and settings it's exclusively referred as Zeus.

addons/editor/initKeybinds.sqf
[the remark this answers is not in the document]
mharis001: > > I dislike the side effect of moving the curator logic object. I guess if we want to avoid that, we would need to add our own ping system.
> 
> I mean, if not this, what else could the location of the curator logic object be used for?

Currently, ALiVE mod causes problems as that updates the curator logic object's position every second. Not our fault but I think it is best to avoid this requirement.

I added an alternative approach that also improves the visuals by making the ping grow and fade out.

mharis001:
I added a custom ping animation that grows the icon and fades it out near the end as well as being unique per curator and displaying their name. Also, using our own logic for viewing the ping (moving camera), which is mostly the same as base game for positions but different for objects. Finally, switched back to "Zeus" instead of "Curators" in the stringtable but not sure about the plural for "Zeus" so opted for "Zeus curators" in the description.

@ampersand38 and @Kexanone please review and merge when ready.

mharis001 (approved):

addons/editor/functions/fnc_handleKeyDown.sqf
```diff
@@ -18,6 +18,50 @@
 
 params ["_display", "_keyCode"];
 
+if (_keyCode in actionKeys "curatorPingView" && {!isNil QGVAR(pingTarget)} && {isNil QGVAR(pingViewed)}) exitWith {
```
rautamiekka, line 21: If anything, the nil checks should be together before checking the key, but I think there's a more worthy aspect to look at, that being dropping the lazy eval in this case cuz if a new context really has a 15% performance penalty, such simple checks like nil (that check things that won't be used in the same statement, no less) should be cheaper than the context creation => lazy eval is more trouble than worth.
```suggestion
if (_keyCode in actionKeys "curatorPingView" && !isNil QGVAR(pingTarget) && isNil QGVAR(pingViewed)) exitWith {
```

addons/editor/functions/fnc_pingCurators.sqf
```diff
@@ -0,0 +1,101 @@
...
+ * None
+ *
+ * Example:
+ * [_curator, _target] call zen_editor_fnc_pingCurators
+ *
+ * Public: No
+ */
+
+#define PING_DRAW_TIME 1.2
+#define PING_DRAW_ICON "\a3\ui_f_curator\data\logos\arma3_curator_eye_512_ca.paa"
```
rautamiekka, line 21: Capitalization just in case:
```suggestion
#define PING_DRAW_ICON "\a3\ui_f_curator\Data\Logos\arma3_curator_eye_512_ca.paa"
```

rautamiekka (changes requested):
Everything I found.

addons/editor/functions/fnc_pingCurators.sqf
mharis001: Does not matter and we generally use full lowercase path (from unpacked P-drive).

addons/editor/functions/fnc_handleKeyDown.sqf
mharis001: In ZEN code style, we always use lazy eval for conditions.

rautamiekka: Frankly, that's a bad excuse.

addons/editor/functions/fnc_pingCurators.sqf
rautamiekka: Doesn't matter even on Linux ?

mharis001: This would be a path in Arma's virtual file system.

rautamiekka: Right. Good to know.

PR #733, opened 2023-04-23
Set negative draw priority for Building Markers

ampersand38:
- Prevent Building Markers from covering player placed markers
- Add setMarkerDrawPriority event
- Call event after building marker creation
- image

### DIFF
--- a/addons/building_markers/functions/fnc_set.sqf
+++ b/addons/building_markers/functions/fnc_set.sqf
@@ -61,6 +61,8 @@ if (_set) then {
     _marker setMarkerDir getDir _object;
     _object setVariable [QGVAR(marker), _marker, true];
 
+    [QEGVAR(common,setMarkerDrawPriority), [_marker, -1]] call CBA_fnc_globalEvent;
+
     // Delete marker when the object is deleted
     private _eventID = _object addEventHandler ["Deleted", {
         params ["_object"];
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -304,6 +304,11 @@
     _object setObjectScale _scale;
 }] call CBA_fnc_addEventHandler;
 
+[QGVAR(setMarkerDrawPriority), {
+    params ["_markerName", "_priority"];
+    _markerName setMarkerDrawPriority _priority;
+}] call CBA_fnc_addEventHandler;
+
 [QGVAR(setVehicleRadar), {
     params ["_vehicle", "_mode"];
     _vehicle setVehicleRadar _mode;


addons/building_markers/functions/fnc_set.sqf
```diff
@@ -61,6 +61,8 @@ if (_set) then {
     _marker setMarkerDir getDir _object;
     _object setVariable [QGVAR(marker), _marker, true];
 
+    [QEGVAR(common,setMarkerDrawPriority), [_marker, -1]] call CBA_fnc_globalEvent;
```
mharis001, line 64: How does this work for JIP?

ampersand38: Ah! Good call. Definitely did not work JIP.
Fix works on testing with loopback clients.

addons/building_markers/functions/fnc_set.sqf
```diff
@@ -61,6 +61,8 @@ if (_set) then {
     _marker setMarkerDir getDir _object;
     _object setVariable [QGVAR(marker), _marker, true];
 
+    [QEGVAR(common,setMarkerDrawPriority), [_marker, -1], _marker] call CBA_fnc_globalEventJIP;
```
mharis001, line 64: Need to handle removing the JIP event if the marker is deleted.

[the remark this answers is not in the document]
ampersand38: Very cool indeed. It would probably need to be `_object` not `_marker` I think, looking at the params?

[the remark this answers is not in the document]
rautamiekka: Maybe, I dunno.

addons/building_markers/functions/fnc_set.sqf
```diff
@@ -61,6 +61,9 @@ if (_set) then {
     _marker setMarkerDir getDir _object;
     _object setVariable [QGVAR(marker), _marker, true];
 
+    private _jipID = [QEGVAR(common,setMarkerDrawPriority), [_marker, -1], _marker] call CBA_fnc_globalEventJIP;
+    [_jipID, _object] call CBA_fnc_removeGlobalEventJIP;
```
mharis001, line 65: Does not handle the marker being removed without the object being deleted.

addons/building_markers/functions/fnc_set.sqf
johnb432: Should building markers be removable for other reasons than building deletion?
Do building markers remain global or should local changes be permitted?

If not, I imagine it's possible to replace the removed marker by detecting when a marker has been deleted with [this EH](https://community.bistudio.com/wiki/Arma_3:_Mission_Event_Handlers#MarkerDeleted):
- You could add this EH on all clients and they would raise an event on the server which would handle the deletion or replacement of the marker, depending if the marker was deleted because the building was deleted or for another reason.
- You could add this EH on the server only and have it deal with the marker deletion or replacement. This assumes that all building marker deletions would be global though.

If building markers should be removable, I guess the EH linked above still can be used, but it would be a bit more tricky: What if `deleteMarkerLocal` is used on building markers? Should the marker be replaced, so that you force people to use `deleteMarker` instead, or should the marker be editable on each client individually, as in permit different marker settings (existence, visibility, size, direction, etc)?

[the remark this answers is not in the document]
ampersand38: [2297b56](https://github.com/zen-mod/ZEN/pull/733/commits/2297b565a0ee0f3905440886c0e30b5e10805426)

addons/building_markers/functions/fnc_set.sqf
```diff
@@ -60,12 +60,16 @@ if (_set) then {
...
+
+    [QEGVAR(common,setMarkerDrawPriority), [_marker, -1], _marker] call CBA_fnc_globalEventJIP;
 
     // Delete marker when the object is deleted
     private _eventID = _object addEventHandler ["Deleted", {
         params ["_object"];
 
-        deleteMarker (_object getVariable [QGVAR(marker), ""]);
+        private _marker = _object getVariable [QGVAR(marker), ""];
+        deleteMarker _marker;
```
johnb432, line 72: Why this change?

I understand if you would want to add a condition, such as this:
```suggestion
        private _marker = _object getVariable [QGVAR(marker), ""];
        
        if (_marker != "") then {
            deleteMarker _marker;
        };
```

Otherwise not really worth it imo.

addons/building_markers/XEH_postInit.sqf
```diff
@@ -2,6 +2,14 @@
 
 if (isServer) then {
     [QGVAR(set), LINKFUNC(set)] call CBA_fnc_addEventHandler;
+
+    addMissionEventHandler ["MarkerDeleted", {
+        params ["_marker"];
+
+        [_marker] call CBA_fnc_removeGlobalEventJIP;
+        [missionNamespace getVariable [_marker, objNull], false] call zen_building_markers_fnc_set;
```
johnb432, line 10:
```suggestion
        [missionNamespace getVariable [_marker, objNull], false] call FUNC(set);
```

addons/building_markers/functions/fnc_set.sqf
johnb432: > nice one [3ea5c51](https://github.com/zen-mod/ZEN/pull/733/commits/3ea5c510db6156c575a069fb72dd9f61dd2b24c7) I don't think any mod can be so protective of its stuff. What use case is there for building markers showing differently between clients?

I agree that a mod shouldn't be that overprotective. As for the use case of building markers showing differently on clients, I don't really know, I'm not a mission maker. I was mostly thinking of how badly the marker system could/would break if you started messing with markers on a per client basis.

addons/building_markers/functions/fnc_set.sqf
```diff
@@ -60,12 +60,16 @@ if (_set) then {
     _marker setMarkerSizeLocal _size;
     _marker setMarkerDir getDir _object;
     _object setVariable [QGVAR(marker), _marker, true];
+    missionNamespace setVariable [_marker, _object];
```
johnb432, line 63: I really would use the marker value in a "raw" format. I'm certain there are mods out there that already do that.

I would use something like this:
```suggestion
    missionNamespace setVariable [format [QGVAR(object_), _marker], _object];
```

I haven't checked if it works, but it would be safer to use imo.

ampersand38: I should have just used netID.
_marker is already `format [QGVAR(%1), _object call BIS_fnc_netId]` `"zen_building_markers_0:1"`.
incidentally,
`allMapMarkers apply {isNil {missionnamespace getVariable _x}} // [true,true]`

addons/building_markers/XEH_postInit.sqf
```diff
@@ -2,6 +2,13 @@
 
 if (isServer) then {
     [QGVAR(set), LINKFUNC(set)] call CBA_fnc_addEventHandler;
+
+    addMissionEventHandler ["MarkerDeleted", {
+        params ["_marker"];
+
+        [_marker] call CBA_fnc_removeGlobalEventJIP;
```
johnb432, line 9: Some checks need to happen before it gets here (i.e filter out all non-ZEN markers).

addons/building_markers/XEH_postInit.sqf
```diff
@@ -2,6 +2,13 @@
 
 if (isServer) then {
     [QGVAR(set), LINKFUNC(set)] call CBA_fnc_addEventHandler;
+
+    addMissionEventHandler ["MarkerDeleted", {
+        params ["_marker"];
+
```
johnb432, line 8:
```suggestion

    // Filter all non-ZEN markers
    if !(QUOTE(ADDON) in _marker) exitWith {};

```

addons/building_markers/XEH_postInit.sqf
```diff
@@ -2,6 +2,16 @@
...
+    // Filter all non-ZEN markers
+    if !(QUOTE(ADDON) in _marker) exitWith {};
```
rautamiekka, line 10:
```suggestion
        // Filter all non-ZEN markers
        if !(QUOTE(ADDON) in _marker) exitWith {};
```

rautamiekka (changes requested):
+1 level of indentation.

PR #736, opened 2023-05-16
Area Markers - Add Editable Markers setting

### LINKED ISSUE #735: Add CBA Setting for Marker Editing Functionality
Yes.
The ability to `Support editing all area markers` #676 has lead to the issue that when playing missions/mods like Antistasi, Vindicta, Liberation or else the markers within those missions, which in most cases are essential for the mission running, can be moved, edited and deleted as well.
Hence it can easily happen that people playing those missions/mods and using ZEN could by accident change/delete one or multiple and therefore break their game.

**Solution you'd like:**
Implement a CBA setting which either is disabled by default or at least can be disabled by code within the mod/mission itself.
This way the feature in itself stays available (certainly a good feature and no reason to remove it) but gives creators/players the option to disable it for "safety" reasons. It's better to be able to prevent mishaps in the first place.

Unfortunately I don't have another easy to implement solution which would give the same functionality and keep the feature within ZEN.

**Additional context:**
If you want to replicate the usecase, subscribe to the Antstasi Mod (https://steamcommunity.com/sharedfiles/filedetails/?id=2867537125), launch it in local hosted with CBA and ZEN loaded and observe that you can edit/delete the markers from Antistasi itseld.

Thank you for your mod.

mharis001:
- title, Close #735


ampersand38 (approved):
Nice one!

PR #742, opened 2023-09-07
Attributes - Add support for ACE Repair 'Enable' setting

LinkIsGrim:
- Title.

See https://github.com/acemod/ACE3/pull/8070, setting requires mission restart so should be fine.

### DIFF
--- a/addons/attributes/initAttributes.sqf
+++ b/addons/attributes/initAttributes.sqf
@@ -953,7 +953,7 @@ if (isClass (configFile >> "CfgPatches" >> "ace_medical")) then {
     ] call FUNC(addAttribute);
 };
 
-if (isClass (configFile >> "CfgPatches" >> "ace_repair")) then {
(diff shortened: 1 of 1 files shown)


addons/attributes/initAttributes.sqf
```diff
@@ -953,7 +953,7 @@ if (isClass (configFile >> "CfgPatches" >> "ace_medical")) then {
     ] call FUNC(addAttribute);
 };
 
-if (isClass (configFile >> "CfgPatches" >> "ace_repair")) then {
+if (isClass (configFile >> "CfgPatches" >> "ace_repair") && {missionNamespace getVariable ["ace_repair_enabled", false]}) then {
```
mharis001, line 956: This runs at preInit where settings aren't guaranteed. Likely need to use the condition parameter of the `addAttribute` function.

LinkIsGrim:
Seems to be working correctly.

PR #743, opened 2023-09-08
Conditionally Load Compats

Mike-MF:
- Uses `skipWhenMissingDependencies` to conditionally load compats from `requiredAddons`
- Merges ACE/Advanced Rappelling compats into main addons
- Adds Advanced Rappelling `requiredAddon`


mharis001 (approved):

Kexanone:
One disadvantage is that this approach won't work when Advanced Rappelling is only present as a server side mod, but I would say that is a small price to play for getting rid of compats.

mharis001:
Because the Advanced Rappelling compat only adds the waypoint and the waypoint's condition will prevent it from being shown when the mod is not loaded, we do not need to use `skipWhenMissingDependencies` for it. This should allow us to support the situation where the Advanced Rappelling mod is loaded only as a server side mod.

mharis001 (approved):

Kexanone:
lgtm, just needs to be tested. I can have a look on my test server when I got time.

PR #748, opened 2023-10-05
Fix Hide Terrain Objects module for tall objects

ampersand38:
- Fix Hide Terrain Objects module needing large radius to capture tall objects

### DIFF
--- a/addons/modules/XEH_postInit.sqf
+++ b/addons/modules/XEH_postInit.sqf
@@ -18,7 +18,7 @@ if (isServer) then {
 
         {
             _x hideObjectGlobal _hide;
-        } forEach nearestTerrainObjects [_position, _objectTypes, _radius];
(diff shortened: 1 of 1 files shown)


Kexanone (approved):

addons/modules/XEH_postInit.sqf
```diff
@@ -18,7 +18,7 @@ if (isServer) then {
 
         {
             _x hideObjectGlobal _hide;
-        } forEach nearestTerrainObjects [_position, _objectTypes, _radius];
+        } forEach nearestTerrainObjects [_position, _objectTypes, _radius, false, true]; // Sort false, 2Dmode true for tall objects
```
mharis001, line 21:
```suggestion
        } forEach nearestTerrainObjects [_position, _objectTypes, _radius, false, true]; // Use 2D mode to handle tall objects
```

mharis001 (approved):

PR #749, opened 2023-10-07
Allow Attach Flag module to be used with flagpoles

Apricot-ale:
- Removes the "is vehicle" check in the attachFlag module function.

It will be possible to attach flags to units (soldiers) and objects (flagpoles).
and it will also be possible to change the flag on the flagpole.
This idea came to me because I was unable to attach a flag from ZEUS to the new flagpole added in 2.14.
They still have to be done from command, which is very inconvenient. This will solve it.


mharis001 (approved):

mharis001:
We still want the check but now also allow flagpoles to be edited using the module.

PR #752, opened 2023-10-09
Improve attach objects module

ampersand38:
- Attach to selection in parent object
- Maintain relative position
- Maintain relative orientation or match selection orientation
- https://www.youtube.com/watch?v=RjuEQJxmY-o

### DIFF
--- a/addons/attached_objects/XEH_PREP.hpp
+++ b/addons/attached_objects/XEH_PREP.hpp
@@ -1,4 +1,5 @@
 PREP(attach);
+PREP(attachToSelection);
 PREP(detach);
 PREP(handleObjectEdited);
 PREP(module);
--- a/addons/attached_objects/functions/fnc_attach.sqf
+++ b/addons/attached_objects/functions/fnc_attach.sqf
@@ -32,12 +32,45 @@
 
     _objects deleteAt _index;
 
-    {
-        _x attachTo [_entity];
+    if (isNil QGVAR(modelSelections)) then {
+        GVAR(modelSelections) = createHashMap;
+    };
 
-        private _dirAndUp = [_entity vectorWorldToModel vectorDir _x, _entity vectorWorldToModel vectorUp _x];
-        [QEGVAR(common,setVectorDirAndUp), [_x, _dirAndUp], _x] call CBA_fnc_targetEvent;
-    } forEach _objects;
+    private _selections = GVAR(modelSelections) getOrDefaultCall [getText (configOf _entity >> "model"), {
+            [""] + (_entity selectionNames LOD_MEMORY select {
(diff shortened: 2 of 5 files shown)


mharis001:
I think maybe this should be a separate keybind. I think it would be annoying to go through the dialog every time you want to attach an object.

ampersand38:
> I think maybe this should be a separate keybind. I think it would be annoying to go through the dialog every time you want to attach an object.

Can do. I'm leaning toward "Attach to Bone" since the description mentions "selected objects", which could get confused with "Attach to Selection".
What behaviour do you want for the attach module?

ampersand38:
And what do you think about attempting to translate from Czech names? https://community.bistudio.com/wiki/ArmA:_Armed_Assault:_Selection_Translations

addons/attached_objects/functions/fnc_attachBone.sqf
```diff
@@ -0,0 +1,83 @@
...
+ * Public: No
+ */
+
+[_this, {
+    params ["_successful", "_objects"];
+
+    if (!_successful) exitWith {};
+
+    curatorMouseOver params ["_type", "_entity"];
+    if (_type != "OBJECT") exitWith {};
```
rautamiekka, line 25:
```suggestion
    if (_type isEqualTo "OBJECT") exitWith {};
```

addons/attached_objects/functions/fnc_attachBone.sqf
```diff
@@ -0,0 +1,83 @@
...
+
+        {
+            if (_selection == "") then {
+                private _dirAndUp = [_entity vectorWorldToModel vectorDir _x, _entity vectorWorldToModel vectorUp _x];
+                _x attachTo [_entity];
+                [QEGVAR(common,setVectorDirAndUp), [_x, _dirAndUp], _x] call CBA_fnc_targetEvent;
+            } else {
+                [_x, _entity, _selection, _isRelative] call FUNC(attachToSelection);
+            };
+        } forEach _objects params ["_selectionPos", "_selectionY", "_selectionZ"];
```
rautamiekka, line 71: Is this actually the intention ?

^ Since `_objects` will be iterated before `params` gets to it, and `params` returns bool based on the success of, or use of a default value in, the creation of the mentioned vars, so even putting it as `forEach (...)` would give an error, IIUC, and the mentioned vars are only used after the `forEach`.

rautamiekka (changes requested):

addons/attached_objects/functions/fnc_attachBone.sqf
ampersand38: Ah, yeah you're right. `forEach` does return the last iteration, but it's not a good use for it, and it was in the wrong place relative to the if anyway.

mharis001:
I think the module can have the "full" attach to options. If it is doable, translating the names would be good, similarly to how hitpoint names are handled.

PR #753, opened 2023-10-11
Add Supply Cargo Attributes

ampersand38:
- Add attributes for supply vehicles to adjust their supply cargo

### DIFF
--- a/addons/attributes/stringtable.xml
+++ b/addons/attributes/stringtable.xml
@@ -17,9 +17,6 @@
             <Chinesesimp>Zeus Enhanced - 属性</Chinesesimp>
             <Turkish>Zeus Enhanced - Özellikler</Turkish>
         </Key>
-        <Key ID="STR_ZEN_Attributes_AmmoCargo2">
-            <English>Ammo Cargo</English>
-        </Key>
         <Key ID="STR_ZEN_Attributes_AmmoCargo">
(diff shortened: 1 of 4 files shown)


addons/attributes/stringtable.xml
```diff
@@ -17,6 +17,18 @@
...
+            <English>Ammo Cargo</English>
+        </Key>
+        <Key ID="STR_ZEN_Attributes_FuelCargo">
+            <English>Fuel Cargo</English>
+        </Key>
+        <Key ID="STR_ZEN_Attributes_RepairCargo">
+            <English>Repair Cargo</English>
+        </Key>
+        <Key ID="STR_ZEN_Attributes_SupportCargo_Tooltip">
+            <English>Cargo to supply to other vehicles.</English>
```
mharis001, line 30: Is this tooltip necessary?

ampersand38: Can certainly be removed.

mharis001:
Also, should add settings to enable/disable these like other attributes.

addons/attributes/stringtable.xml
mharis001: It sounds a bit too generic. I guess we can either remove or add more specific tooltips.

mharis001 (approved):
Code LGTM.

PR #755, opened 2023-11-14
Add chatter keybind

### LINKED ISSUE #324: FEATURE REQUEST: Chatter Keybind & Dialog Saving Option
It would be nice for the Chatter Keybind introduced in Achilles 1.3.0 to be merged with ZEN as well. Also, the fact that the chatter dialog saves the last inputted message even if you select another character is sometimes counter-intuitive, especially in cases where you can't afford to waste the few seconds selecting all the text, removing it, and typing your new message.

**Solution you'd like:**
- A keybind to open the chatter module on a selected unit.
- Prevent the dialog from saving the last inputted message, regardless of character, either by addon option or outright removal.

- Make chatter module accessible through the context menu, but that may clutter the already-busy unit context menus.

### LINKED ISSUE #719: Use Chatter module while remote controlling an AI unit.
**Is your enhancement related to a problem?**
Opening the chat while remote controlling AI opens up the regular chat box.

**Solution you'd like:**
When remote controlling AI and I open the chat, I'd like it to use the Chatter module instead of chatting as myself.

N/A

**Additional context:**
Not completely sure, but I believe this was a feature in Achilles.

ampersand38:
- Add keybind to open chatter dialog for remote controlled or selected unit
- Close #719 
- Close #324

### DIFF
--- a/addons/editor/initKeybinds.sqf
+++ b/addons/editor/initKeybinds.sqf
@@ -105,6 +105,16 @@
     };
 }, {}, [0, [false, false, false]]] call CBA_fnc_addKeybind; // Default: Unbound
 
+[ELSTRING(main,DisplayName), QGVAR(ModuleChatter), ELSTRING(modules,ModuleChatter), {
+    if (isRemoteControlling player) exitWith {
+        [remoteControlled player] call EFUNC(modules,moduleChatterUnit);
+    };
+
+    if (isNull curatorCamera) exitWith {};
+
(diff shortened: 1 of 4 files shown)


mharis001:
Maybe we should extract this out to a separate component. That way the editor component does not depend on the modules component.

Also, maybe we should override the chat keybind when remote controlling to instead use the chatter functionality. Not sure about the UX on that however (perhaps a setting would be useful here).

PR #756, opened 2023-12-17
Added: Server events for getting the code string, for logging in missions

killerswin2:
Add a server event that mods and mission makers can listen to. This was done in the case of wanting to log what code was being executed on objects, players, groups... etc. In our case for Antistasi, we want to be able to look at our log file to see if someone is executing code through Zen. 

- objectExecCode event:
params:
_code - string of code to execute
_mode - code execution mode
_entity - object entity that refers to _this 

- groupExecCode event:
params:
_code - string of code to execute
_mode - code execution mode
_entity - group entity that refers to _this 

- guiExecCode
_code - string of code to execute
_mode - code execution mode
_args - arguments passed to the code

### DIFF
--- a/addons/attributes/initAttributes.sqf
+++ b/addons/attributes/initAttributes.sqf
@@ -375,6 +375,7 @@
     [QGVAR(objectExecHistory), QGVAR(objectExecMode), LSTRING(ExecObject_Tooltip), 20, 1000],
     {
         _value params ["_code", "_mode"];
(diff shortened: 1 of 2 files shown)


veteran29:
As this is executed in UI it feels to me like these should be local events. If you want to log on the server you could handle them and forward to server for logging.

killerswin2:
yeah that's right, they should be local events

PR #765, opened 2024-06-07
Area Markers - Only allow editing when in markers mode

### LINKED ISSUE #731: Only allow manipulating area markers when in marker tab
**Is your enhancement related to a problem?**
Currently if the map is open all area markers can be moved, deleted and edited even those created in 3DEN. This can lead to accidentally moving or deleting these markers when handling units or objects on the map.

**Solution you'd like:**
Only allow moving, deleting and editing area markers when the user is in the marker creation tab. The editing icon for each area marker is only shown when in the marker creation tab. 

Alternatively, the ability to manipulate area markers could be toggled with a keybind.

**Additional context:**
This is what I mean by marker creation tab:  
image

mharis001:
- Only allow editing area markers when in markers mode (matches behaviour of engine-based icon markers)
    - Close #731 
- Switch back to PFH for updating icon visibility
    - Event based method for detecting map visibility does not always work (e.g., when a dialog is opened)


addons/area_markers/functions/fnc_isInEditMode.sqf
```diff
@@ -0,0 +1,21 @@
...
+ * In Marker Editing Mode <BOOL>
+ *
+ * Example:
+ * [] call zen_area_markers_fnc_isInEditMode
+ *
+ * Public: No
+ */
+
+visibleMap
+&& {!dialog}
```
rautamiekka, line 19: Simple bool checks are too cheap for lazy eval, I even tested it myself.
```suggestion
&& !dialog
```

rautamiekka (changes requested):

addons/area_markers/functions/fnc_isInEditMode.sqf
mharis001: In ZEN, we always use lazy eval as a matter of style. The difference in performance, if any, is negligible.

rautamiekka: A selfish pure excuse, as invalid as a null check on a nil variable, as dangerous as going down the Niagara Falls in a barrel.

addons/area_markers/functions/fnc_isInEditMode.sqf
```diff
@@ -0,0 +1,21 @@
...
+&& {!call EFUNC(common,isInScreenshotMode)}
+&& {RscDisplayCurator_sections select 0 == 3}
```
Kexanone, line 21: We should probably define the different tabs with preprocessors

Kexanone (changes requested):

addons/area_markers/functions/fnc_isInEditMode.sqf
mharis001: Will fix in rest of the code base as well.

Kexanone (approved):

PR #785, opened 2024-12-26
Show 3DEN comments in Zeus

Timi007:
Hello everyone! I would like to donate a feature from one of my mods. 

- Shows 3DEN comments in Zeus 3D and on the map

### How it works
All comments get stored in as a scenario attribute that is available during mission runtime. Currently, only the comment name is displayed and the description/tooltip is ignored. For it to work, the mission must be saved before preview. You can also hide an 3DEN comment by including the directive `#ZEN_IGNORE#` in the description/tooltip of the comment.

zen_3den_comments_zeus

### TODO

- [ ] Documentation

I also plan on adding the ability to create comments as Zeus during mission runtime. I think a module and context menu action would be nice. How do you handle internal dependencies? Should the module and action be implemented in their respective addons `modules` and `context_actions`? Or should they be implemented in this addon and depend on `modules` and `context_menu` for the classes and API?

Furthermore, I cannot store the comments without saving (I thought I could do it with the new `OnBeforeMissionPreview` event, but it does not work for me). And displaying tooltips on 3D icons is also not possible. So if anyone has some idea on how to solve these problems, I would be grateful.

### DIFF
--- a/addons/comments/initSettings.inc.sqf
+++ b/addons/comments/initSettings.inc.sqf
@@ -1,0 +1,17 @@
+[
+    QGVAR(enable3DENComments),
+    "CHECKBOX",
+    [LLSTRING(3DENComments), LLSTRING(3DENComments_Description)],
+    [ELSTRING(main,DisplayName), LSTRING(DisplayName)],
+    true,
+    0
+] call CBA_fnc_addSetting;
+
+[
+    QGVAR(3DENCommentsColor),
+    "COLOR",
+    [LLSTRING(3DENCommentsColor), LLSTRING(3DENCommentsColor_Description)],
+    [ELSTRING(main,DisplayName), LSTRING(DisplayName)],
+    [0, 1, 0.75, 1],
+    0
+] call CBA_fnc_addSetting;
--- a/addons/comments/stringtable.xml
+++ b/addons/comments/stringtable.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project name="ZEN">
+    <Package name="Comments">
+        <Key ID="STR_ZEN_Comments_DisplayName">
+            <English>Comments</English>
+            <German>Kommentare</German>
+        </Key>
+    </Package>
+    <Package name="3DENComments">
+        <Key ID="STR_ZEN_Comments_3DENComments">
+            <English>Show 3DEN comments in Zeus</English>
+            <German>Zeige 3DEN-Kommentare im Zeus</German>
+        </Key>
+        <Key ID="STR_ZEN_Comments_3DENComments_Description">
+            <English>Show 3DEN comments in Zeus.</English>
+            <German>Zeige 3DEN-Kommentare im Zeus an.</German>
+        </Key>
+        <Key ID="STR_ZEN_Comments_3DENCommentsColor">
+            <English>3DEN comments color</English>
+            <German>Farbe der 3DEN-Kommentare</German>
+        </Key>
+        <Key ID="STR_ZEN_Comments_3DENCommentsColor_Description">
+            <English>Color of 3DEN Kommentare in Zeus.</English>
+            <German>Farbe der 3DEN-Kommentare im Zeus.</German>
+        </Key>
+    </Package>
+</Project>
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -50,6 +50,7 @@ PlayerBot
 Ryan180602
 shukari
 SilentSpike
+Timi007
 Toinane
 Tuupertunut
 veteran29
--- a/addons/comments/$PBOPREFIX$
+++ b/addons/comments/$PBOPREFIX$
@@ -1,0 +1,1 @@
+x\zen\addons\comments
--- a/addons/comments/Cfg3DEN.hpp
+++ b/addons/comments/Cfg3DEN.hpp
@@ -1,0 +1,32 @@
+class Cfg3DEN {
+    class Attributes {
+        class Default;
+        class GVAR(HiddenAttribute): Default {
+            onLoad = QUOTE((ctrlParentControlsGroup ctrlParentControlsGroup (_this select 0)) ctrlShow false);
+        };
+    };
+
+    class Mission {
+        class Scenario {
+            class AttributeCategories {
+                class PREFIX {
+                    collapsed = 1;
+                    displayName = ECSTRING(main,DisplayName);
+
+                    class Attributes {
+                        class GVAR(3DENComments) {
+                            property = QGVAR(3DENComments);
+                            value = 0;
+                            control = QGVAR(HiddenAttribute);
+                            displayName = CSTRING(DisplayName);
+                            tooltip = "";
+                            defaultValue = "[]";
+                            expression = "";
+                            wikiType = "[[Array]]";
+                        };
+                    };
+                };
+            };
+        };
+    };
+};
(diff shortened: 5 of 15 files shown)


mharis001:
You should implement the modules/context menu actions in this component. In terms of tooltips, maybe you could try using 2D icons similarly to how the `doors` component works.

addons/comments/initSettings.inc.sqf
```diff
@@ -0,0 +1,70 @@
+[
+    QGVAR(enableComments),
```
mharis001, line 2: `QGVAR(comment)`

addons/comments/initSettings.inc.sqf
```diff
@@ -0,0 +1,70 @@
...
+        TRACE_1("Enable comments setting changed",_this);
+
+        private _display = findDisplay IDD_RSCDISPLAYCURATOR;
+        if (!_this || GVAR(enable3DENComments) || isNull _display || GVAR(draw3DAdded)) exitWith {};
+        [_display] call FUNC(addDrawEventHandler);
+    }
+] call CBA_fnc_addSetting;
+
+[
+    QGVAR(commentColor),
```
mharis001, line 18: `QGVAR(colors)`

addons/comments/initSettings.inc.sqf
```diff
@@ -0,0 +1,70 @@
...
+    {
+        TRACE_1("Comment color setting changed",_this);
+        params ["_r", "_g", "_b", "_a"];
+
+        GVAR(commentsActiveColor) = [_r, _g, _b, 1];
+    }
+] call CBA_fnc_addSetting;
+
+[
+    QGVAR(enable3DENComments),
```
mharis001, line 33: I think setting could be reworked as "Import 3DEN Comments" and potentially merged with the allow deletion one: "No", "Yes", "Yes + Prevent Deletion".

addons/comments/stringtable.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project name="ZEN">
+    <Package name="Comments">
```
mharis001, line 3: Just use one Package.

mharis001 (commented):
Just some initial style comments.

What is the reason for not using the GUI controls for drawing the icons as well? Not suggesting that is the best approach but curious on whether its just for visuals?

Timi007:
> Just some initial style comments.
> 
> What is the reason for not using the GUI controls for drawing the icons as well? Not suggesting that is the best approach but curious on whether its just for visuals?

The icons are GUI controls. Do you want to turn the text (comment title and creator) into GUI controls? If so, I thought about it. I'll need to try it out first, as I'm not sure how it will look visually with the scaling at distance. So for now I'm using `drawIcon` for text because it seemes easier for me. I will try to make a prototype that only uses GUI controls.

This is how it currently looks (left 3D, right map):
zen_comments

addons/comments/stringtable.xml
```diff
@@ -0,0 +1,63 @@
...
+            <Italian>Commenti</Italian>
+            <Spanish>Comentarios</Spanish>
+            <French>Commentaires</French>
+            <Chinese>評論</Chinese>
+            <Korean>댓글</Korean>
+            <Japanese>コメント</Japanese>
+            <Portuguese>Comentários</Portuguese>
+            <Chinesesimp>评论</Chinesesimp>
+            <Turkish>Yorumlar</Turkish>
+        </Key>
```
Timi007, line 19: This is the same as `STR_3DEN_Comment_textPlural`. Should I remove this entry and use `STR_3DEN_Comment_textPlural` for the display name?

addons/comments/initSettings.inc.sqf
Timi007: This setting does not prevent placing comments. It only determines if to show or to hide them. So maybe `QGVAR(showComments)`?

mharis001: I think the setting should be "enabled" or not. To me, it does not make sense to be able to place comments if I cannot see them.

addons/comments/stringtable.xml
mharis001: Yes.

mharis001:
Just a couple things to consider:
- Using a HashMap to store the comments and storing both the 3DEN and Zeus created ones in one data structure.
- Editing comments by double clicking them.
- Allowing color to be set per comment. Not sure if this is bloat or unnecessary.
- Wish list: Moving comments (this would be quite involved).

As for the text scaling, I would let's try to mimic the behaviour of 3DEN.

addons/comments/initSettings.inc.sqf
Timi007: I thought about it and I must say I don't like this. As a client you might want to ignore 3DEN comments but as a server you want to force to preventimg deletion for other clients that see 3DEN comments.

mharis001:
Does this handle a 3DEN comment being deleted by a client and then later another client connects? Is that comment also deleted for the JIP client?

mharis001:
This has some similarities with the `area_markers` component. I think using some of the approaches found in that could be useful. For example, how creating, deleting, and updating markers works as well as how those changes are reflected in the editable icons.

Timi007:
> Just a couple things to consider:
> 
> * Using a HashMap to store the comments and storing both the 3DEN and Zeus created ones in one data structure.
> * Editing comments by double clicking them.
> * Do we need the module? 3DEN just uses the context menu for placing comments right?
> * Allowing color to be set per comment. Not sure if this is bloat or unnecessary.
> * Wish list: Moving comments (this would be quite involved).
> 
> As for the text scaling, I would say let's try to mimic the behaviour of 3DEN.

* Will do
* Will do
* That's right. In hindsight, maybe we don't need the module.
* I'm also unsure. But I think it would be nice. It allows users to "categorize" their comments. Maybe a dropdown with color presets is more user friendly instead of a color picker?
* Will consider it but for now low priority
* IIRC 3DEN doesn't scale comments at distance but just hides them after 750m. The icon scaling (but not text) is a Zeus thing. Currently, I've combined it to: Hide after 750m (3DEN), don't scale text (3DEN/Zeus), icon scales (Zeus)

> Does this handle a 3DEN comment being deleted by a client and then later another client connects? Is that comment also deleted for the JIP client?

No, I think it doesn't. Good catch. I will do a MP test when thr PR is close to merging.

> This has some similarities with the `area_markers` component. I think using some of the approaches found in that could be useful. For example, how creating, deleting, and updating markers works as well as how those changes are reflected in the editable icons.

Yeah, I already try to implement this similar to that component.

mharis001:
Tried it out in-game. Works pretty well in terms of the visuals👍.

I messed around with some of the refactoring/cleanup I commented about earlier but ran out of time. Here is the diff in case you want to get a better idea of what I meant: https://github.com/Timi007/ZEN/compare/comments...zen-mod:ZEN:comments

Timi007:
> Tried it out in-game. Works pretty well in terms of the visuals👍.
> 
> I messed around with some of the refactoring/cleanup I commented about earlier but ran out of time. Here is the diff in case you want to get a better idea of what I meant: [Timi007/ZEN@comments...zen-mod:ZEN:comments](https://github.com/Timi007/ZEN/compare/comments...zen-mod:ZEN:comments)
> 
> Pretty much what you have, mainly just using a HashMap and code style similar to the rest of the code base.

Oh, I just pushed a commit with my changed including editing comments. I will incorporate your code style into my changes.

mharis001:
Instead of the ignore string, we could add a 3DEN checkbox attribute to comments for ignoring them.

Timi007:
Do you know how to change the cursor in Zeus? Also, I think a mechanic for locking the position of comments would be nice. Do you have any ideas about how this mechanic could look like?

mharis001:
I don't there is a way to change the cursor. In terms of locking, probably just a checkbox in the dialog.

Timi007:
I think this is ready for review.

BrettMayson (approved):

PR #800, opened 2025-03-06
Add 3D and map plotting tools

Timi007:
- Add 3D and map plotting tools

This feature is meant to help the user measure distances and directions. The start and end positions can be either static or attached to an object, where the values are then updated dynamically. The user can toggle between different distance, speed, and azimuth formats. More formatters can be defined in `class zen_plotting_formatters`. The color can be changed in the CBA settings. Plotting is possible in 3D or on the map.

https://github.com/user-attachments/assets/da3b7885-be33-4ed6-9975-f1339fd9c100

### TODO
- [x] Fix line connections of rectangular cuboid
- [x] Fix attaching plot to objects
- [x] ~~Optimize drawing (do not draw if outside screen)~~ The overhead of checking if the positions are in view is too high ~2x perf decrease
- [x] Add icons
- [x] ~~Handle text rendering when attaching multiple plots to objects~~ Limited to one end point for a unit at a time
- [x] Fix key handlers not always register input
- [x] Add the time it takes to travel the distance depending on a speed


addons/plotting/initKeybinds.inc.sqf
```diff
@@ -0,0 +1,61 @@
...
+    {
+        if (!isNull curatorCamera && {!GETMVAR(RscDisplayCurator_search,false)}) then {
+            private _numFormatters = count GVAR(distanceFormatters);
+            GVAR(currentDistanceFormatter) = (GVAR(currentDistanceFormatter) + 1) % _numFormatters;
+
+            true
+        };
+    },
+    {},
+    [DIK_R, [false, false, false]] // Default: R
```
mharis001, line 16: These keybinds conflict with the same base game Zeus controls. I think a default behind some modifiers since this doesn't need to be so accessible would be good.

addons/plotting/CfgFormatters.hpp
```diff
@@ -0,0 +1,30 @@
...
+            priority = 90;
+        };
+        class Yards {
+            formatter = QUOTE(format [ARR_2('%1 yd',(_value * 1.094) toFixed 1)]);
+            priority = 80;
+        };
+        class Mile {
+            formatter = QUOTE(format [ARR_2('%1 mi',(_value / 1609.344) toFixed 2)]);
+            priority = 70;
+        };
```
mharis001, line 18: I think these could be made more complex. For example, I don't think the decimal point is needed after 100 m and for any moving object it just flickers. Essentially, increase precision for small values, decrease for large ones.

addons/plotting/functions/fnc_compileFormatters.sqf
```diff
@@ -0,0 +1,52 @@
...
+
+    private _formatters = [];
+    {
+        private _entryConfig = _x;
+
+        private _formatterName = configName _entryConfig;
+
+        private _formatterString = getText (_entryConfig >> "formatter");
+        if (_formatterString isNotEqualTo "") then {
+            _formatterString = "params ['_value']; " + _formatterString;
```
mharis001, line 29: Unnecessary, just pass the value in `_this` and then, `format ["%1 m", _this]` for example.

mharis001 (commented):
Interesting functionality. Just some initial thoughts.

Timi007:
Thanks for the initial feedback! Can you maybe provide feedback on following question too?
1. Currently, the max distance the lines and icons are rendered is hardcoded to 3km. Should this be a setting or maybe linked to the view distance?
2. What measurements units should be present? As a user of the metric system, do we need feet, yards, and miles?
3. For the line plot, I also want to add the time it takes to travel the distance depending on a speed. Should this speed be also a toggle like distance and azimuth formats? Or a setting? Or toggle + setting?
4. When attaching a plot point to an object, the formatted texts overlap. In 3D, we can solve this using the text Y offset. On the map there is no parameter. Should only the text of the last plot be shown? Or do we just restrict attaching one plot to one object?

addons/plotting/CfgFormatters.hpp
Timi007: Good idea

mharis001:
1. I think just a reasonable max value is fine. For example, the `visibility` component uses a max of 5 km.
2. You could combine it to "systems" instead of "units". So, metric: meters -> kilometers depending on distance, and imperial: feet -> yards -> miles. Not sure.
3. Probably just a toggle.
4. Even with multiple plots with offset text, it would be difficult to tell which one belongs to which plot. Again, not sure what the best UX here is.

addons/plotting/functions/fnc_drawLine.sqf
```diff
@@ -0,0 +1,63 @@
...
+ * Public: No
+ */
+
+params ["_startPos", "_endPos", "_visualProperties", "_formatters", ["_ctrlMap", controlNull, [controlNull]]];
+_visualProperties params ["_icon", "_color", "_scale", "_angle", "_lineWidth"];
+
+private _distance = _startPos vectorDistance _endPos;
+
+private _azimuthToStart = _endPos getDir _startPos;
+private _azimuthToEnd = _startPos getDir _endPos;
```
mharis001, line 36: Should this be the relative direction?

Timi007: This is the direction from start to end pos and end to start. Both `_startPos` and `_endPos` are PosASL. I don't understand the relative part in your question.

mharis001: IIRC, when attached to objects the direction value being displayed did not seem intuitive. If that makes sense.

Timi007: Need to check what you mean. It should show the same value as the unit just opening their compass.

mharis001: I guess it just comes to whether we want to know that the position is 90 degrees relative to the vehicle or just the difference in their bearings. The former what what I assumed the value was.

Timi007: What purpose does relative bearing w.r.t. the attached unit have to the Zeus or any unit on the ground? I understand if you want the direction as a relative clock, but I have never given or received relative bearing. For artillery corrections, you also don't use relative directions.

addons/plotting/initKeybinds.inc.sqf
Timi007: Are these better? 
- <kbd>Ctrl</kbd>+<kbd>R</kbd>: Switch Distance
- <kbd>Shift</kbd>+<kbd>R</kbd>: Switch Speed
- <kbd>Ctrl</kbd>+<kbd>T</kbd>: Switch Azimuth

Timi007:
Should we only allow attaching one plot at a time for now? I honestly don't have a better solution.

BrettMayson (changes requested):
Doing `Alt + R` didn't change the speed format when I tested

addons/plotting/initKeybinds.inc.sqf
```diff
@@ -0,0 +1,99 @@
...
+                        TRACE_2("Automatically removed next invalid plot",_lastPlotId,_lastPlot);
+                    };
+                };
+            };
+
+            true
+        };
+    },
+    {},
+    [DIK_Y, [false, true, false]] // Default: Ctrl + Y
```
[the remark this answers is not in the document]
Timi007: In the German keyboard layout, the Z and Y keys are switched. Meaning Ctrl + Y for German keyboards is Ctrl + Z. 

I don't know how to handle the case that standard keybind in CBA should be Ctrl + Z and for German keyboards Ctrl + Y (because Arma automatically converts Y → Z).

Timi007:
> Doing `Alt + R` didn't change the speed format when I tested

Currently, I only implemented km/h as speed format. The toggle switch is there if anyone wants to add a custom speed format. I don't know if it makes sense to add mph or knots as speed formats?

addons/plotting/stringtable.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project name="ZEN">
+    <Package name="Plotting">
+        <Key ID="STR_ZEN_Plotting_DisplayName">
+            <English>Plotting</English>
+            <German>Plotten</German>
+        </Key>
+        <Key ID="STR_ZEN_Plotting_MeasureDistance">
+            <English>Measure distance</English>
```
mharis001: These strings should all be title case for consistency with everything else.

addons/plotting/script_component.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include "\x\zen\addons\main\script_macros.hpp"
+
+#include "\a3\ui_f\hpp\defineDIKCodes.inc"
+#include "\x\zen\addons\common\defineResinclDesign.inc"
+
+#define ICON "\a3\ui_f\data\map\markerbrushes\cross_ca.paa"
+#define ICON_ANGLE 0
+#define ICON_SCALE 1
+#define MAP_ICON_SCALE 24
+#define LINEWIDTH 5
```
mharis001: `LINE_WIDTH`.

BrettMayson:
I thought I had seen MPH, but I was mistaken, so makes sense for just KM/H

addons/plotting/stringtable.xml
Timi007: Fixed

addons/plotting/script_component.hpp
Timi007: Changed

BrettMayson (approved):

PR #809, opened 2026-05-23
Add listenable setDamage event when Attribute set

Alien314:
- Adds a setDamage event for mod/mission script compatibility to attributes/initAttributes.inc.sqf and context_actions fnc_repairVehicles.sqf

### DIFF
--- a/addons/context_actions/functions/fnc_repairVehicles.sqf
+++ b/addons/context_actions/functions/fnc_repairVehicles.sqf
@@ -24,4 +24,5 @@ private _vehicles = _this select {
 
 {
     _x setDamage 0;
+    [QEGVAR(common,setDamage), [_x, _damage], _x] call CBA_fnc_targetEvent;
 } forEach _vehicles;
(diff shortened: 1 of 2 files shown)


mharis001:
`setDamage` is GA/GE. A local event probably makes the most sense (consumers can rebroadcast if needed). Also, the naming of this should be considered - should this be namespaced under `common`?

Alien314:
Nearly every other event uses common, so I thought it was simplest since setDamage isn't necessarily an attribute specific thing. Edit: added event to context_actions fnc_repairVehicles as well.

addons/context_actions/functions/fnc_repairVehicles.sqf
```diff
@@ -24,4 +24,5 @@ private _vehicles = _this select {
 
 {
     _x setDamage 0;
+    [QEGVAR(common,setDamage), [_x, _damage], _x] call CBA_fnc_targetEvent;
```
mharis001: `_damage` is undefined here.

mharis001:
The `common` events are generally meant to be internal use only.

In either case, what is the intended use case for this? If we do implement this, then there are a lot more places where `setDamage` is called and we would want to consider those as well.

Alien314:
Mainly for resetting flags or effects related to the damage state of vehicles.

PR #811, opened 2026-07-07
Inventory - Preserve nested container contents when editing

Venrix:
- Preserve the contents of containers (e.g. backpacks) inside a vehicle when its inventory is edited via the 'edit inventory' dialog
- Fix items inside those containers being lost on confirm (see video)

https://github.com/user-attachments/assets/ff3017e7-dec5-4678-8701-7051c5dba244

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -53,5 +53,6 @@ SilentSpike
 Timi007
 Toinane
 Tuupertunut
+Venrix
 veteran29
 vivngh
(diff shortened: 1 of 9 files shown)


mharis001:
The current implementation restores container contents by matching only on container class. This breaks if a user removes a container and then adds a new container of the same class before confirming: the new container is indistinguishable from the old one, so it incorrectly receives the old container's inventory.

We need to track container instances/lifecycle while the dialog is open, so only containers that actually survived the edit have their previous contents restored.

mharis001:
I've pushed an improved implementation that should better handle these edge cases.

Also, project convention is that PBO authors array is reserved for original authors or those that contributed substantially to the given module.

mharis001 (approved):

mharis001:
Tested in-game. Second set of eyes would be good though.

PR #812, opened 2026-08-21
Add object alive state filter to Update Editable Objects module

UnseenKill:
- add to the **"UPDATE EDITABLE OBJECTS"** dialog a filter to select whether objects should be alive, dead or both.

image


mharis001:
I did some minor cleanup and made some improvements. I think the option is best placed at the end instead of between curators and range. Also, changed "both" to "either, which makes more sense to me since an object can't both be dead and alive at the same time.

image

mharis001 (approved):

UnseenKill:
> I did some minor cleanup and made some improvements. I think the option is best placed at the end instead of between curators and range. Also, changed "both" to "either, which makes more sense to me since an object can't both be dead and alive at the same time.
> image

I wasn't braining correctly when choosing to call it "both". Thanks for considering the PR.

UnseenKill:
People might complain about "dead" being the default option, now. Care if I switch options to "Either", "Yes", "No" instead?

mharis001:
Oversight on my part. Yes, default should be either. Good catch.

mharis001 (approved):

